PR #157, opened 2014-12-11
Added builders for `use ` & `namespace` nodes

kix:


nikic:
A "bit" late, but the use builder is now in as well: https://github.com/nikic/PHP-Parser/commit/0c34706799d298864a8e0cd4b1cb3eff710540bc With an `as` method for a custom alias.

kix:
@nikic, well, that's great; I missed the `as` method, thanks for adding it.

PR #163, opened 2015-01-09
Create .gitattributes with basic export-ignores

rmruano:
Files and directories with the attribute export-ignore won’t be added to archive files. This will prevent composer from downloading unnecessary files (i.e: tests) when --prefer-dist is set.


TomasVotruba:
Great!

PR #166, opened 2015-01-14
Added getMethods to Interface_ and Trait_.

tomzx:


nikic:
To avoid copying this into the three classes and allow simpler `instanceof` checks I've split this off into a `ClassLike` parent, see https://github.com/nikic/PHP-Parser/commit/1366e833a16c4267883187da6da7b317f0582dd7.

tomzx:
I like it, thanks @nikic!

PR #175, opened 2015-02-13
prevent a unnecessary concat

staabm:
analog to https://github.com/nikic/PHP-Parser/commit/b7ffdbbb929f9f07780a236659d1d1baf68ef137

### DIFF
--- a/lib/PhpParser/NodeDumper.php
+++ b/lib/PhpParser/NodeDumper.php
@@ -21,7 +21,7 @@ class NodeDumper
         }
 
         foreach ($node as $key => $value) {
-            $r .= "\n" . '    ' . $key . ': ';
+            $r .= "\n    " . $key . ': ';
 
             if (null === $value) {
                 $r .= 'null';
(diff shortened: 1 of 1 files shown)


nikic:
Just to clarify, the change in https://github.com/nikic/PHP-Parser/commit/b7ffdbbb929f9f07780a236659d1d1baf68ef137 targets a particular case where `$r` is a huge string and doing one concat instead of two (which also means one realloc instead of two) makes a large difference. In the two cases that changed here it doesn't make a practical difference.

I've merged it anyway, if only to have consistent code in one file :)

staabm:
Your argument also counts for https://github.com/nikic/PHP-Parser/pull/175/files#diff-195da172622beb6ed449385e53e1f145L24 doesnt it? Its the same $r and now 1 alloc instead of 2...

The other line isnt that important, I agree.

nikic:
Nope, the concat there happens before the concat with `$r`, so it is operating on a small string only.

PR #176, opened 2015-02-14
Added php shebang

fonsecas72:
So I don't have to ```php bin/php-parse.php``` and just ```bin/php-parse.php``` instead.


nikic:
Sorry for the delay, now merged as https://github.com/nikic/PHP-Parser/commit/e3bc5b564d2eaceae8ecb74ff8194efa6235a04f :)

PR #177, opened 2015-02-25
Improved BuilderFactoryTest ccv

fonsecas72:
Hi, I've been playing around with the tests so I can feel this package better and I've found that there were some methods not covered by them and I started by just doing that.
I thought you might find this useful, but if not, it's ok anyway :)

If you're interested I can also improve this PR. E.g. split this big integration test into several others.


nikic:
Sorry for the delay, I was out of town.

Apart from the static property, these additions don't seem to be testing the builder functionality, but rather test manual creation of nodes (which builders are there to avoid). So if we want to test this, this probably isn't the right place. I'm not sure if it has to be tested as its implicitly tested by the parser tests.

Btw, the `new Expr\Variable("value['sasa']")` instantiation in particular should be `new Expr\ArrayDimFetch(new Expr\Variable('value'), new Scalar\String('sasa'))`.

PR #180, opened 2015-03-02
Implicit public should not generate as explicit public

thomasruiz:
So I ran into a little problem: the PrettyPrinter cannot generate methods without visibility. Since it could be used sometimes (e.g. generating specs for [phpspec](https://github.com/phpspec/phpspec)), it should be available.

I added some tests and fixed others to make sure it works properly, but maybe I missed some use cases... Let me know what you think.


nikic:
Current behavior was implemented in https://github.com/nikic/PHP-Parser/commit/0f69f12b94900396c71754bdabef945e6907688e. Your suggestion makes sense as it differentiates between implicit/explicit and at the same time preserves `isPublic()` behavior. I've merged this as https://github.com/nikic/PHP-Parser/commit/73cace360d676ce4284aa7e19a07b4e39c3d3eda with a tweak to preserve the previous behavior for the property builder - it will never generate obsolete `var` property declarations.

PR #191, opened 2015-04-03
Autoloader::register() has side effects even when the library is not used

malkusch:
Hi,
Do you by any chance remember the intention behind this in `Autoloader::register()` (83a2077f):

    ini_set('unserialize_callback_func', 'spl_autoload_call')

Well, it has side effects even if the library is not used. Composer will register the autoloader and break under some circumstances unit tests which don't use the library at all. I created one simple test case to present the issue. If you would run `AutoloaderTest::testRunInSeparateProcess()` with the original autoloader it would break. The funny thing, any project which just pulls in PHP-Parser can't have such unit tests.
I suggest to simply remove that line from the autoloader or at least don't turn that switch on by default.

### DIFF
--- a/lib/PhpParser/Autoloader.php
+++ b/lib/PhpParser/Autoloader.php
@@ -23,7 +23,6 @@ class Autoloader
             return;
         }
 
-        ini_set('unserialize_callback_func', 'spl_autoload_call');
         spl_autoload_register(array(__CLASS__, 'autoload'), true, $prepend);
         self::$registered = true;
         self::$runningOnPhp7 = version_compare(PHP_VERSION, '7.0-dev', '>=');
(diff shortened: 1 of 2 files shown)


nikic:
IIRC I originally copied the autoloader over from Twig, which at the time included this `ini_set` call. I just checked and it looks like they dropped this [about two years ago](https://github.com/twigphp/Twig/commit/536ea37d1f87a882ce8a5d62680caf881ecaf6fe).

So yeah, I'd say this serves no purpose whatsoever and can be dropped.

nikic:
Merged the second commit via https://github.com/nikic/PHP-Parser/commit/8d18848fb0e119f1a66089c75f459e3212b8dabd. I had to drop the process isolation test because, for whatever weird reasons, PhpStorm goes totally crazy when it tries to run that test (though running phpunit from cli works)...

malkusch:
Thank you.

PR #194, opened 2015-04-14
added column-numbers to syntax errors

addiks:
Hello,

this fork add's column-numbers to detected syntax errors.

Example:
```
Syntax error, unexpected T_STATIC, expecting T_STRING on line 1, column 16 to 22
```

I need this because i am currently writing a plugin for my text-editor that should display PHP syntax errors. For that i need a way to not only get the line-number but also the column numbers of an error to show more accurately where the error is. Currently there is no practical usable command line tool that can just lint one file alone and tell me not only the line in which a syntax-error is but also the column-numbers. I need this, please integrate it into the main repository, thank you.

### DIFF
--- a/lib/PhpParser/Error.php
+++ b/lib/PhpParser/Error.php
@@ -6,6 +6,7 @@ class Error extends \RuntimeException
 {
     protected $rawMessage;
     protected $rawLine;
+    protected $columns;
 
     /**
      * Creates an Exception signifying a parse error.
@@ -13,9 +14,10 @@ class Error extends \RuntimeException
      * @param string $message Error message
      * @param int    $line    Error line in PHP file
      */
-    public function __construct($message, $line = -1) {
+    public function __construct($message, $line = -1, array $columns=array()) {
         $this->rawMessage = (string) $message;
         $this->rawLine    = (int) $line;
+        $this->columns    = $columns;
         $this->updateMessage();
     }
 
@@ -67,6 +69,13 @@ class Error extends \RuntimeException
             $this->message .= ' on unknown line';
         } else {
             $this->message .= ' on line ' . $this->rawLine;
+        }
+
+        if(isset($this->columns[0]) && isset($this->columns[1])){
+            $this->message .= ', column ' . $this->columns[0];
+            if($this->columns[1] > $this->columns[0]){
+                $this->message .= ' to ' . $this->columns[1];
+            }
         }
     }
 }
--- a/lib/PhpParser/Error.php
+++ b/lib/PhpParser/Error.php
@@ -6,18 +6,23 @@ class Error extends \RuntimeException
 {
     protected $rawMessage;
     protected $rawLine;
-    protected $columns;
+    protected $tokens = array();
+    protected $tokenIndex;
+    protected $beginColumnCache;
 
     /**
      * Creates an Exception signifying a parse error.
      *
-     * @param string $message Error message
-     * @param int    $line    Error line in PHP file
+     * @param string $message    Error message
+     * @param int    $line       Error line in PHP file
+     * @param array  $tokens     Array of all tokens in the file that caused the error.
+     * @param int    $tokenIndex Index in $tokens of token where error happened.
      */
-    public function __construct($message, $line = -1, array $columns=array()) {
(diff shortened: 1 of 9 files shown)


lib/PhpParser/Error.php
```diff
@@ -13,9 +14,10 @@ class Error extends \RuntimeException
      * @param string $message Error message
      * @param int    $line    Error line in PHP file
      */
-    public function __construct($message, $line = -1) {
+    public function __construct($message, $line = -1, array $columns=array()) {
```
GrahamCampbell: cs

lib/PhpParser/Error.php
```diff
@@ -68,5 +70,12 @@ protected function updateMessage() {
         } else {
             $this->message .= ' on line ' . $this->rawLine;
         }
+
+        if(isset($this->columns[0]) && isset($this->columns[1])){
```
GrahamCampbell: cs

lib/PhpParser/Error.php
```diff
@@ -68,5 +70,12 @@ protected function updateMessage() {
...
+            $this->message .= ', column ' . $this->columns[0];
+            if($this->columns[1] > $this->columns[0]){
```
GrahamCampbell: cs

nikic:
I don't think column numbers should be added to errors by default (there are a couple of projects that wouldn't like that, at least). However we should definitely have the ability to get this info.

I think the best way to do this would be to provide the `Error` object with the complete attributes of the token (or node) where the error occurred. This should allow easily computing the column numbers based on `startFilePos` and `endFilePos` (or the token position as done here, either would work). This would also allow having column information to errors not thrown directly from the parser. (And would also provide the end line).

addiks:
I have now added a parameter (--with-column-info / -c) to the php-parse.php script that will show column-information for syntax-errors, made the error-class work with token-data instead of columns directly and written a test for the new behaviour in the error-class. The default behaviour does not change anymore but needs to be triggered using that parameter.

nikic:
Merged in https://github.com/nikic/PHP-Parser/commit/33a39fae06dec77ba262fae61b2727fe960d5215 with some more tweaks in https://github.com/nikic/PHP-Parser/compare/33a39fae06dec77ba262fae61b2727fe960d5215...62f83a0dc25e191b369b3c77415059c263009a23.

There are a few open issues:
 * What columns should be used for EOF errors? Right now these won't have column info.
 * Should column info be added to errors that don't have accurate position information? E.g. if the best that can be done is mark an entire class as an error?
 * Should columns be 0-based or 1-based? Currently they're 0-based, but 1-based seems to be more common for columns.

addiks:
I think EOF errors should display the last character as the error to directly show where the error is without interfering too much. When errors would mark an entire class (or such) as faulty it should only show the class-name at the top as an error to prevent other errors in that class to be hidden by the whole class to be marked as an error. In cases where no detailed column-informations are available it is fine to not have/show column-information, just adding artificial column-numbers would create indistinguishable situations. Column-numbers should be 1-based to be consistent with line-numbers. (I didn't implemented that 0-based on purpose, i just build it so that it lined up correctly in my text-editor.) That are however not hard opinions. If you want to build it differently just do it, i am fine with it.

nikic:
Column numbers switched to 1-based in https://github.com/nikic/PHP-Parser/commit/c8454271e10008d34e4fdc0b9d2e17d72834a931.

nikic:
Added info for EOF errors in https://github.com/nikic/PHP-Parser/commit/a35c2a20671199fffcb424d9183f0a959b105065. I've put the location at one past the end of the file. Might be better to put it one past the end of the last non-whitespace, non-comment token instead - phpstorm does this for example.

Anyway, closing this issue as the feature should be mostly complete now.

felixfbecker:
I find it a bit weird that column numbers are 1-based while line numbers are 0-based. At least this should be documented.

nikic:
@felixfbecker It is documented both in phpdoc (https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Error.php#L96) and and the error documentation (https://github.com/nikic/PHP-Parser/blob/master/doc/component/Error.markdown#column-information)

felixfbecker:
Sorry I noticed that seconds after I wrote my comment.

PR #206, opened 2015-06-25
added syntax highlighting in section php code

royopa:


nikic:
Merged via https://github.com/nikic/PHP-Parser/commit/7fbbf83011fe6e2564d54b720270f963408e08e5. It's not PHP code, but the highlighting happens to work nicely ^^

PR #209, opened 2015-07-05
Correct detection of implicitly public method nodes

lisachenko:
Method nodes that have only one of `abstract`, `final` or `static` are not marked as public. Method should be considered public if it has explicit `public` modifier or it isn't private or protected.


nikic:
Nice catch, merged. The same issue also existed for static properties, so this is fixed now as well.

PR #218, opened 2015-09-14
PhpParser - create must be a static fn

ovr:
Why is it not a static fn?


nikic:
The suggested usage for this method is `(new ParserFactory)->create(ParserFactory::PREFER_PHP7)` rather than `ParserFactors::create(ParserFactory::PREFER_PHP7)`. The latter is a perfectly reasonable alternative in this case (as it's unlikely this class will ever have a meaningful constructor), but I try to avoid static methods unless strictly necessary.

PR #221, opened 2015-09-14
Use PSR4 autoload system by composer

ovr:
- [x] PSR4 by Composer
- [x] Use phpunit by composer
- [x] TravisCI - run phpunit with debug and composer install


ovr:
ping @nikic 
It's needed for next scrutinizer ci, because i am going to setup "sending code coverage report" to it

.travis.yml
```diff
@@ -11,3 +11,10 @@ matrix:
   allow_failures:
     - php: 7.0
   fast_finish: true
+
+install:
+  - composer self-update
```
GrahamCampbell: :-1:

.travis.yml
```diff
@@ -11,3 +11,10 @@ matrix:
...
+  - composer --prefer-source install
```
GrahamCampbell: `composer install --prefer-source`

.travis.yml
```diff
@@ -11,3 +11,10 @@ matrix:
...
+
+script:
+  - ./vendor/bin/phpunit -v --debug
```
GrahamCampbell: missing newline

.travis.yml
fezfez: @GrahamCampbell why ?

GrahamCampbell: why add it?

GrahamCampbell: just wastes time

ovr:
@GrahamCampbell 
Omg, schoolchild, go away with this spaming atack

AveVlad:
> Omg, schoolchild, go away with this spaming atack

:+1: @GrahamCampbell  go contirube laravel

.travis.yml
fezfez: Until composer does not have a stable release, its need. By the way its pretty fast

.travis.yml
ovr: @GrahamCampbell 
It's not a standart to write parameter in the end
For example, On OSX Just try to use cp with `-R parameter` in the of exec comand

.travis.yml
```diff
@@ -11,3 +11,11 @@ matrix:
...
+
```
GrahamCampbell: too many newlines

.travis.yml
GrahamCampbell: it is pretty standard for symfony, which is what composer uses

.travis.yml
GrahamCampbell: It's not needed though. If you removed it, everything would still work.

GrahamCampbell:
> Omg, schoolchild, go away with this spaming atack

No need to be rude. The way I see it, I'm preventing bad changes being made to a project I care about.

.travis.yml
fezfez: what do mean ?

GrahamCampbell: you've added 2 newline characters instead of 1

ovr: 1 new line in the of file, what is wrong?

.travis.yml
ovr: > pretty standard for symfony
Where did you read it?

.travis.yml
ovr: @GrahamCampbell 
Sometimes ago I got fails without it
It's not slow

GrahamCampbell: meh, can we at least move it to "before-install" then

fezfez: @GrahamCampbell : Good solution

nikic:
Switched to the PSR-4 autoloader in https://github.com/nikic/PHP-Parser/commit/e3a93561780f750489aa3a1cf99d23dcfa44520f. We may need to switch back to a custom one later if there's going to be aliases again, but I guess it's better to use the generated one for now.

.travis.yml
ovr: Guys, it's rly useless but it's needed 1 line for before install and 1 line empty line after, 2 uneeded lines

nikic:
Phpunit dep in composer is now present as well: https://github.com/nikic/PHP-Parser/commit/6f3fd7834a60e176d7734b786cfd1df6b764ed5e

PR #222, opened 2015-09-14
Phpunit - show awesome code coverage in console

ovr:


GrahamCampbell:
Why not just use --coverage-text on travis?

ovr:
@GrahamCampbell 
This will works for all envs

GrahamCampbell:
> This will works for all envs

Yeh, I know. I'm questioning why we actually want that.

ovr:
It's a suggestion, in my opinion "it's rly cool"

GrahamCampbell:
I don't really like it turned on by default, and it's not, for good reason.

nikic:
Agree with @GrahamCampbell, coverage shouldn't be on by default. It would be nice to get travis to send coverage to scrutinizer for some version (5.6 makes most sense probably?)

ovr:
I will close it, and send PR with scrutinizer CI code coverage

PR #226, opened 2015-09-17
Travis are not building php 7.1

GrahamCampbell:
https://github.com/travis-ci/php-src-builder/commit/17395d3214936af2ac92ccecc806e99b8863c332.

Nightly is still 7.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,9 @@ php:
   - 5.5
   - 5.6
   - 7.0
-  - nightly
   - hhvm
 
 install: composer install
 
 matrix:
-  allow_failures:
-    - php: nightly
   fast_finish: true


nikic:
I realize nightly is still on 7.0 now, but it should be switched over sometime soon, right? Is there a problem with keeping it in the list?

GrahamCampbell:
>  but it should be switched over sometime soon, right?

No, they intentionally made nightly 7.0., but you'd have to ask them what they plan on doing once php 7.0  is actually stable, because I'm not sure.

nikic:
https://github.com/travis-ci/php-src-builder/commit/17395d3214936af2ac92ccecc806e99b8863c332 has been reverted by https://github.com/travis-ci/php-src-builder/commit/b147483b691cfcfe6c25ab8d297b46e043a2a04c a couple of days ago, so `nightly` now points to PHP 7.1 :)

Issue #225, opened 2015-09-14
Use ScrutinizerCI


ovr:
Can you setup https://scrutinizer-ci.com/?
After it I will send PR with configuration for check? okey?

dupuchba:
:+1: for this one

nikic:
The project is on scrutinizer under https://scrutinizer-ci.com/g/nikic/PHP-Parser but I couldn't figure out how to add a GH service hook.

GrahamCampbell:
Scrutinizer should add it for you. Maybe disable and re-enable the project on scrutinizer, or contact support to get it added again.

GrahamCampbell:
Shame they don't support repairing the webhook. The service I work on supports that.

ovr:
@nikic 
Did you try to remove and add it again?

GrahamCampbell:
> Did you try to remove and add it again?

I literally just said that:

> Maybe disable and re-enable the project on scrutinizer

ovr:
@GrahamCampbell 
I didnt read you messages

> Shame they don't support repairing the webhook. The service I work on supports that.

How is it related to this issue?

GrahamCampbell:
I was just saying that it's possible, and easy to implement, since I've implemented it before. ;)

nikic:
I've removed/added the project and now the hook is there, so should be possible to test configs now.

ovr:
@nikic Super, I will send PR in few hours

nikic:
I've decided to not use Scrutinizer. It spews up a lot of notices but only very little is of interest to me (out of a couple dozen messages I looked at two were genuinely useful). I've set up coveralls for the test coverage, as that was the most interesting part of Scrutinizer.

PR #229, opened 2015-09-21
[WIP] Scrutinizer CI

ovr:


.travis.yml
```diff
@@ -17,4 +17,8 @@ matrix:
     - php: nightly
   fast_finish: true
 
-script: vendor/bin/phpunit
\ No newline at end of file
+script: vendor/bin/phpunit --coverage-clover=coverage.clover
```
GrahamCampbell: you should only generate coverage on 5.6, not every version

nikic:
Heh, it looks like Scrutinizer doesn't support reopen to run another inspection :/

I'm wondering whether using Scrutinizer is really worthwhile or we should instead use coveralls for the coverage (which is the only part of Scrutinizer I really care about).

nikic:
Closing per #225.

PR #230, opened 2015-09-21
Remove test_old runner

ovr:


nikic:
This one is still in use. Should maybe rename it to something other than test_old...

ovr:
Okey, but can We move it to test dir and rename to runner?

PR #232, opened 2015-10-01
More pretty print tests

AydinHassan:
Hey, I was using the `PrinterPrinter` for something and noticed that a lot of the code was not tested, so I decided to add some tests. See 818cc6.

I also added a multi init for loop parser test. See 2188234.


test/code/prettyPrinter/for.test
```diff
@@ -0,0 +1,28 @@
...
+
+for ($i = 0; $i < 10; $i++) {
+
+}
+
+for ($i = 0,$j = 0; $i < 10; $i++) {
+
+}
+
+for ($i = 0; $ < 10;) {
```
nikic: The `$` here is missing a variable name -- that's probably what's causing the issue.

The error message in the test is really bad though, that should be done better...

AydinHassan: Wow, that's embarrassing :(

AydinHassan:
@nikic Fixed and updated the PR description, now excuse me while I go back to page 1 in my PHP book :sob:

nikic:
Thanks for the tests! That brings pretty printer coverage to 91% :)

AydinHassan:
Thanks :smile:

PR #148, opened 2014-11-17
Add ScopeVisitor

ockham:
Following discussion in #144

Issue #196, opened 2015-04-30
Scope chain


edsonmedina:
Currently there's no way of knowing the scope of the current node (unless your Visitor is manually tracking it).

IMO it would be extremely handy to have a public property in NodeVisitorAbstract with a the scope chain.

So visitors could do something like:

    if ($node instanceof ClassMethod)
    {
        // we've got a method, now find its class name
        foreach ($this->scopeChain as $scope)
        {
            if ($scope instanceof Class_)
            {
                $className = $class->name;
            }
        }
        
        // ...
    }

This would be very easy to implement.

nikic:
I don't think this belongs in `NodeVisitorAbstract` (which only job is to provide you empty method bodies). You can easily implement a separate visitor providing the context information that is necessary for your use case.

edsonmedina:
I imagine you're describing something like this?

    $context = new ContextStack(); 
    
    // this visitor manages the stack
    $traverser->addVisitor (new ContextVisitor ($context)); 
    
    // this visitor consumes the stack
    $traverser->addVisitor (new SomeOtherVisitor ($context)); 

But nothing guarantees me the `ContextVisitor` will always be the first to run, right?

I usually make individual visitors keep track of all the relevant context they need, but this gets verbose quickly (lots of `instanceof` and state) and happens frequently.

It would be so much simpler if the context was provided by the parser (it already has the context in memory).

It could be an optional component if you think this would create unnecessary overhead.

nikic:
Visitors are always run in the order they're registered. (I.e. they run concurrently, but for each node they are run in-order, which should be sufficient for managing context.)

sekjun9878:
Please see https://github.com/nikic/PHP-Parser/pull/234

edsonmedina:
Nikic's answer is satisfactory enough for me.

A visitor can track the scope and others can have a reference to it (or to a shared context object).

Closing this.

PR #234, opened 2015-10-23
Implement ScopeResolver with tests and docs. Fixes #144, Extends #148.

### LINKED ISSUE #144: Question: Get function scope
Hi, nikic. I'm parsing peace of code like this:

``` php
<?php
function myFunc($a) 
{
    return strlen($a);
}
```

And I have visitor like this:

``` php
<?php

use PhpParser\Node;
use PhpParser\Node\Stmt;
use PhpParser\Node\Expr;

class FunctionNodeVisitor extends PhpParser\NodeVisitorAbstract
{
       public function leaveNode(Node $node) {
              // function call
               if ($node instanceof Node\Expr\FuncCall) {
                   // ...... some code
               }
       }
}
```

Is it possible to get "parent"-function (`myFunc`) when I'm in the node `strlen`?
Thanks

sekjun9878:
@nikic 
@ockham 

I've re-implemented #148 with proper differentiation between functions, classes, and Closures, along with corresponding docs and tests.

## Scopes

Root namespace: `\`
Function `test()` inside the root namespace: `\test()`
Raw `Vendor\Package\Tester` namespace: `\Vendor\Package\Tester\`
Function `test()` inside the namespace: `\Vendor\Package\Tester\test()`
Class `TestClass` inside the namespace `\Vendor\Package\Tester\TestClass`
Class method `coverage()`: `\Vendor\Package\Tester\TestClass->coverage()`
A Closure inside the root namespace: `\Closure`
Nested Closures inside the root namespace: `\Closure::Closure::Closure`
A Closure inside namespaced function `test()`: `\Vendor\Package\Tester\test()::Closure::Closure`
A Closure inside namespaced class method `coverage()`: `\Vendor\Package\Tester\TestClass->coverage()::Closure`

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ Component documentation:
 
  1. [Error](doc/component/Error.markdown)
  2. [Lexer](doc/component/Lexer.markdown)
+ 3. [ScopeResolver](doc/component/ScopeResolver.markdown)
 
  [doc_1_x]: https://github.com/nikic/PHP-Parser/tree/1.x/doc
  [doc_master]: https://github.com/nikic/PHP-Parser/tree/master/doc
--- a/doc/component/ScopeResolver.markdown
+++ b/doc/component/ScopeResolver.markdown
@@ -1,0 +1,71 @@
+ScopeResolver documentation
+===========================
+
+ScopeResolver is a default NodeVisitor for adding `scope` attributes to each of the nodes.
+
+Having a scope attribute handily available for every statement can be really useful, for example, when trying to
+figure out when a variable assignment goes out of scope.
+
+Usage
+-----
+
+Using ScopeResolver is extremely simple. There are no options, just add ScopeResolver to NodeTraverser
+and let it run. Here is an example:
+
+```php
+use PhpParser\ParserFactory;
+use PhpParser\NodeTraverser;
+use PhpParser\NodeVisitor\ScopeResolver;
+
+$parser = (new ParserFactory)->create(ParserFactory::PREFER_PHP7);
+$traverser = new NodeTraverser;
+
+$traverser->addVisitor(new ScopeResolver);
+
+try {
+    // read the file that should be converted
+    $code = file_get_contents($file);
+
+    // parse
+    $stmts = $parser->parse($code);
+
+    // traverse
+    $stmts = $traverser->traverse($stmts);
+
+    // see the results!
+    var_dump($stmts);
+} catch (PhpParser\Error $e) {
+    echo 'Parse Error: ', $e->getMessage();
+}
+```
+
+Now each of the nodes should have a `scope` attribute following the syntax form below.
+
+Syntax
+------
+
+In case you require additional information about a scope, it is given to you in an easily parsable string format.
+
+All scopes are prepended by `\`, which is the root namespace.
+
+All namespaces (without anything inside) end with `\` to indicate a raw namespace.
+
+Functions are appended by the function definition braces `()`.
+
+Class methods are prepended by the object operator `->`.
+
+Closures are essentially Closure classes, so they are treated as such in a namespace.  
+If inside a class or a function, or if they are nested, they are prepended by the scope resolution operator `::`.
+
+## Examples:
+
+Root namespace: `\`  
+Function `test()` inside the root namespace: `\test()`  
+Raw `Vendor\Package\Tester` namespace: `\Vendor\Package\Tester\`  
+Function `test()` inside the namespace: `\Vendor\Package\Tester\test()`  
+Class `TestClass` inside the namespace `\Vendor\Package\Tester\TestClass`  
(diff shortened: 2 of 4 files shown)


sekjun9878:
Whoops, put the wrong class name in the commit message. If you could modify that when you apply the patch that'd be fantastic.

nikic:
Same question here as in #148 and #196: Why add a scope string as an attribute to nodes? Apart from debugging, I'm not sure I see why one would be interested in having a scope string in an arbitrary format.

To expand on what was said in the other issues/PRs, here is the usage that I had in mind:

```php
class ScopeResolver {
    // Just an example API
    public function getNamespace() : ?Stmt\Namespace_;
    public function getClass() : ?Stmt\ClassLike;
    public function getMethod() : ?Stmt\ClassMethod;
    public function getFunction() : ?Stmt\Function_;
    //??? public function getFunctionLike() : ?Stmt\FunctionLike;
    public function getScopeString() : string;

    // Maybe also getClassName() etc? Dunno.
}

$traverser = new NodeTraverser;
$scopeResolver = new ScopeResolver;
$traverser->addVisitor(new MyVisitor($scopeResolver));
$traverser->addVisitor(new MyOtherVisitor($scopeResolver));

// Where for example (resolving magic consts)
class MyVisitor extends NodeVisitorAbstract {
    private $context;
    public function __construct(ScopeResolver $context) {
        $this->context = $context;
    }

    public function leaveNode(Node $node) {
        if ($node instanceof MagicConst\Class_) {
            $class = $this->context->getClass();
            return new Scalar\String_($class ? $class->namespacedName : '');
        } else if ($node instanceof MagicConst\Namespace_) {
            $ns = $this->context->getNamespace();
            return new Scalar\String_($ns ? $ns->toString() : '');
        }
        /* resolve other magic consts */
    }
}

// Or (adding the scope string as an attribute)
class MyOtherVisitor extends NodeVisitorAbstract {
    private $context;
    public function __construct(ScopeResolver $context) {
        $this->context = $context;
    }
    public function enterNode(Node $node) {
        $node->setAttribute('scope', $this->context->getScopeString());
    }
}
```

Basically the idea is to provide the context information in a way where you can access the individual parts of the context separately and completely (e.g. you could get not only the name of the function, but also check whether you're currently in a function that returns by reference etc.)

Does that make sense?

sekjun9878:
@nikic Sure, but all you're doing is wrapping the string format into a class and giving that instead. I could write a parser for the format I've described (which would be easy) and it would be the exact same result.

A Scope class is pretty useless for a scope resolver, because at the end of the day, every scope is unique and a string format perfectly represents that. Sort of like a scope hash, except this is more descriptive.

With the string format, I could easily put it into a hash table(array key), along with all the variables and functions inside (the scope), and be guaranteed that the string is a unique external representation of that scope.

Also, I was following the pattern from NameResolver and the previous PR as well, so instead of another Visitor having to implement this Resolver just to use it - I can run it, and have the scope result embedded in the attributes.

nikic:
> Sure, but all you're doing is wrapping the string format into a class and giving that instead. I could write a parser for the format I've described (which would be easy) and it would be the exact same result.

Parsing the format to extract information from it doesn't make a lot of sense: If I want to know the current class in a visitor, it would be simpler to just reimplement the state management than to try parsing this scope string. Even if you do that, it only gives you back a class name, not the AST node of the class.

I think I am currently misunderstanding what this is about. I assumed it's for providing additional commonly needed context to visitors, so you don't have to store the current class, current method etc in every visitor.

From your description this seems to be about something else, you're looking for unique scope keys to use in a dictionary. I'm not sure I completely get the idea behind that, but in any case these keys are not unique for two reasons: Firstly, there can be multiple closures at one scope level, which would collide. Secondly, if you account for conditional definitions it's possible to have two classes/functions/etc in one file with the same name.

sekjun9878:
@nikic You're right, the use case for this isn't to back-reference a class or a function inside a visitor, but rather provide a unique identifier for each Node which can be accessed after the traversal is complete. My use case needs to keep track of function scopes / variables inside scopes (a static type analyser) and a string format is the simplest to deal with, since scopes do not inherit there is no need for a complex data structure.

I guess what could be done is have a Scope object like you've implemented (which will contain the AST Nodes), but have that added as an attribute to each Node and not as an injected complement to another visitor - similar to NameResolver. That way, visitors can access the scope object by getAttribute, it can be accessed after all traversals in the same way, and a __toString() method can convert the Scope to a unique string that can be used in a hash table / and for comparison.

I'm not sure how the Closure collision issue can be dealt with - right now I'm using startFilePos since that is guaranteed to be unique for each closure, but the default lexer doesn't select that as a default option so something alternative might need to be used?

I'm not sure how the conditional function issue can be solved when converting to string format short of a pointer keeping track of functions with the same name, and appending numbers.

PR #241, opened 2015-12-29
Remove some unecessary @property doc comments

jesseschalken:
I don't think these need to be there anymore.


GrahamCampbell:
:+1:

GrahamCampbell:
Indeed. Those docs are only useful document properties that exist through php's magic methods.

nikic:
Yep, those are leftovers from an older implementation using magic properties. Thanks!

PR #246, opened 2016-02-15
Test under PHP5.3

theofidry:


theofidry:
Sorry for the PR, just wanted to check what would happen under PHP 5.3 :/

PR #247, opened 2016-02-15
Better Travis configuration

theofidry:
* Enable Travis cache for Composer
* Update composer before install to keep it up to date
* Authenticate Composer against GitHub to avoid hitting the API limit rate: **requires to add a Travis environment variable `GITHUB_OAUTH_TOKEN` with the token value. The token can be generated in your account panel and all boxes should be unchecked for the permissions right (i.e. do not giving it any permission - only use for authentication)
* Use `--prefer-dist` over `--prefer-source` to download less and make it faster

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ language: php
 
 sudo: false
 
+cache:
+  directories:
+    - $HOME/.composer/cache
+
 php:
   - 5.4
   - 5.5
@@ -12,7 +16,7 @@ php:
 
 install:
   - if [ $TRAVIS_PHP_VERSION = '5.6' ]; then composer require satooshi/php-coveralls '~1.0'; fi
-  - composer install --prefer-source
+  - composer install --prefer-dist
 
 matrix:
   allow_failures:
@@ -24,3 +28,4 @@ script:
 
 after_success:
   if [ $TRAVIS_PHP_VERSION = '5.6' ]; then php vendor/bin/coveralls; fi
+


.travis.yml
```diff
@@ -1,6 +1,9 @@
+sudo: false
```
GrahamCampbell: no need to move this

.travis.yml
```diff
@@ -10,11 +13,16 @@ php:
...
-
 matrix:
   allow_failures:
     - php: nightly
   fast_finish: true
 
-script: vendor/bin/phpunit
\ No newline at end of file
+before_install:
+  - composer self-update
```
GrahamCampbell: self-update is unneeded

theofidry: Is there a point to keep using an old version of composer?

Taluu: just not the concern of this lib

theofidry: Not following you @Taluu: I don't see the point of risking to break the tests because of not having Composer up to date. Besides composer also evolves with time to resolve dependencies faster and in a better way. Why disregarding that?

Taluu: Once again, because this is not the concern of the lib to actually have an updated composer. Nothing should prevent someone with a out-of-date composer installation to install the lib. Otherwise, why limit to composer and not upgrade php and everything else ?

theofidry: >Nothing should prevent someone with a out-of-date composer installation to install the lib.

If that's your concern you should add tests for that to effectively try it with and outdated composer version, not a random version on which you have no control of.

>Otherwise, why limit to composer and not upgrade php and everything else ?

I don't see how it is related here. Upgrading PHP has an influence on the code of your application/library. In this case it doesn't matter if you're using composer v1.0.0-alpha1 or v1.0.0-alpha11, BC or not. It would be a different story if we were talking of a composer extension, but that's not the case here.

Taluu: > In this case **it doesn't matter if you're using composer v1.0.0-alpha1 or v1.0.0-alpha11**, BC or not. It would be a different story if we were talking of a composer extension, but **that's not the case here**.

You said it yourself. It doesn't matter which version of composer is used on travis.

theofidry: So as the Composer version doesn't matter (to the library), why disregarding the benefits brought by newest version?

Taluu: Which are ? Even looking at the requirements, the composer install is almost unecessary...

Taluu: And also : https://docs.travis-ci.com/user/languages/php#Installing-Composer-packages

As it is said, it is perfectly safe to ignore this warning

theofidry: >And also : https://docs.travis-ci.com/user/languages/php#Installing-Composer-packages

Ha *finally*, something more constructive! Indeed if Travis composer version is updated periodically, absolutely no need for that.

Taluu: that's as constructive as the fact that this is not the concern of this lib to have a composer up to date but whatever...

nikic:
It looks like Oauth is no longer necessary as GitHub dropped the rate limit: https://github.com/composer/composer/issues/4884#issuecomment-195229989 Though there's a report further down that the issue still exists...

theofidry:
Indeed, updated @nikic :)

keradus:
:+1:

nikic:
Finally merged as https://github.com/nikic/PHP-Parser/commit/e7ca4b7b04e1e10e37e2e411fc8c47a5bb6c9bae :) However it does not seem like `prefer-dist` is working out: https://travis-ci.org/nikic/PHP-Parser/jobs/116168686#L150

theofidry:
@nikic maybe it just requires another build?

nikic:
@theofidry Yep, everything works fine after triggering a new build. Thanks for the pointer!

theofidry:
You're welcomed :)

PR #252, opened 2016-03-03
Update file mode

stevemiketa:
Some mechanism that I haven't been able to figure out, I'm guessing it's a feature of Composer, is modifying the permissions of `bin/php-parse`. However, this only happens when Composer decides to download the source; installing with --prefer-dist doesn't cause the below issue. After running an initial `composer install` I run `composer install --no-dev` to generate a classmap with no dev dependencies, but it fails due to the modified permissions:

```
$ composer install --prefer-source
Loading composer repositories with package information
Installing dependencies (including require-dev)
  - Installing nikic/php-parser (v2.0.1)
    Cloning ce5be709d59b32dd8a88c80259028759991a4206

Writing lock file
Generating autoload files
$ cd vendor/nikic/php-parser/ && git status && cd ../../..
HEAD detached at v2.0.1
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   bin/php-parse

no changes added to commit (use "git add" and/or "git commit -a")
```

* PHP 5.4.45
* Composer version 1.0-dev (78ffe0fd087e3e2941c7d85b71f294985a8007ad) 2016-01-25 18:34:52

This PR just updates the permissions of `bin/php-parse` to executable.


nikic:
Looks reasonable, merged :)

stevemiketa:
Fantastic.  Thanks for the fast reply!

PR #266, opened 2016-04-13
PrettyPrinter:Unnecessary escape in single quoted.

algo13:
```
>php-parse --pretty-print "<?php echo 'path\to';
====> Code <?php echo 'path\to';
==> Pretty print:
<?php

echo 'path\\to';
```

### DIFF
--- a/lib/PhpParser/PrettyPrinter/Standard.php
+++ b/lib/PhpParser/PrettyPrinter/Standard.php
@@ -98,7 +98,7 @@ class Standard extends PrettyPrinterAbstract
                 }
                 /* break missing intentionally */
             case Scalar\String_::KIND_SINGLE_QUOTED:
-                return '\'' . $this->pNoIndent(addcslashes($node->value, '\'')) . '\'';
+                return '\'' . $this->pNoIndent(addcslashes($node->value, '\'\\')) . '\'';
             case Scalar\String_::KIND_HEREDOC:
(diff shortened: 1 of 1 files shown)


nikic:
Escaping `\` is not unnecessary if you consider something like `'\\\''`. This parses to string contents `\'`, and printing it as `'\\''` would not be correct. This is the cause of the build failure @ https://travis-ci.org/nikic/PHP-Parser/jobs/122780318.

algo13:
sorry.
It was my mistake.

print \0 ?
```
>php-parse --pretty-print "<?php echo trim('str', \"\n\");
#!/usr/bin/env php
====> Code <?php echo trim('str', "\n");
==> Pretty print:
<?php

echo trim('str', "\n");
```
```
>php-parse --pretty-print "<?php echo trim('str', \"\0\");
====> Code <?php echo trim('str', "\0");
==> Pretty print:
<?php

echo trim('str', " "); //< real \0
```

nikic:
Escaping `\0` sounds like a good idea. However we should probably escape the entire `\0-\37` control character range (excluding the existing `\n\r\t\f\v`) and not just `\0`.

nikic:
Merged as https://github.com/nikic/PHP-Parser/commit/371c78334473dd80b6376e0f81faabecb97e4510, extended to the whole control character range.

PR #268, opened 2016-04-15
make autoloader more PSR-4

remicollet:
Minor change to avoid relying on dirname(__DIR__) == PhpParser

This is useful for downstream distribution where we'd like to provides both v1.x and v2.x installed in a different tree


remicollet:
Thanks!

PR #269, opened 2016-04-15
support -h and --help standard options

remicollet:
Could be nice to also have a -V / --version standard option...


nikic:
Merged as https://github.com/nikic/PHP-Parser/commit/d1dd9f5aecbb3fa95b7a1f007572dc980a90a4e6 with an output tweak in https://github.com/nikic/PHP-Parser/commit/82bb6627c9b804e1d40d29fc9ec97a39165f0e8e. Thanks!

PR #277, opened 2016-05-14
Node\Stmt\Foreach_ - fix php doc block

ovr:
Hey!

Wrong PHP docblock

Thanks


nikic:
`foreach ($array as $foo->key => $foo->value) {}` is legal, so keyVar and valueVar are not necessarily simple variables, but may be any variable expression.

ovr:
о0
Thank you answer 👍

Issue #280, opened 2016-06-08
Inline PHP breaks HTML elements


moorscode:
Consider the following source:
```php
foreach ( $tabs as $identifier => $tab ) : ?>
		<a class="nav-tab" id="<?php echo $identifier; ?>-tab" href="#top#<?php echo $identifier; ?>"><?php echo $tab['label']; ?></a>
<?php endforeach;
```

This will be outputted by PrettyPrint in the following manner:
```php
foreach ($tabs as $identifier => $tab) {
    ?>
		<a class="nav-tab" id="<?php 
    echo $identifier;
    ?>
-tab" href="#top#<?php 
    echo $identifier;
    ?>
"><?php 
    echo $tab['label'];
    ?>
</a>
		<?php 
}
```
Because of the extra newlines the HTML will have spaces in places where it wouldn't be in the source.

Thus I was wondering why the following new-line was placed after the closing PHP tag:
https://github.com/nikic/PHP-Parser/blob/371c78334473dd80b6376e0f81faabecb97e4510/lib/PhpParser/PrettyPrinter/Standard.php#L794

I can't come up with a reason why it would have to be there.

moorscode:
A problem here is that new-lines are not being parsed and disappear, so we cannot "restore" them at the proper position.

nikic:
The additional newlines are stripped by PHP -- they will not be part of the output. This is only a question of formatting.

We can add an attribute to InlineHTML that tracks whether or not there was a leading newline in the original code and then reproduce it that way in the pretty printer.

nikic:
Though from a quick look, adding this attribute is not as simple as it might be, as the newline is part of the `T_CLOSE_TAG`, which is no longer available when the InlineHTML is parsed, as it has been reduced as a statement terminator previously.

nikic:
https://github.com/nikic/PHP-Parser/commit/ec614c95dd5fb840be165e4abde0d66c58299b09 adds a new attribute for InlineHTML that specifies whether a newline should/shouldn't be included. This should ensure proper formatting for both cases.

PR #281, opened 2016-06-08
Remove new-line after PHP closing tag

moorscode:
There shouldn't have to be a forced new-line after a closing PHP tag.
The non-PHP code still contains all new-lines it had so adding additional stuff would break original formatting.

Fixes https://github.com/nikic/PHP-Parser/issues/280


nikic:
Resolved via https://github.com/nikic/PHP-Parser/commit/ec614c95dd5fb840be165e4abde0d66c58299b09, which will print the newline depending on whether the original code had it.

Issue #178, opened 2015-02-25
Please provides test


remicollet:
Hi

Since 1.1.0 and 0dd2b9a74c46149847d79e72cf418a0f519e3ed9 tests are no more provided in the downloadable tarball.

Can you please consider removing  the export-ignore line for test in .gitattribute, and provide them again ?

     test export-ignore

TomasVotruba:
@remicollet Hi, this dues to composer optimization, IIRC. You should be able to obtain test via cloning the repository.

remicollet:
Yes, I'm able to get the test suite from a clone of the repo, but this is terribly ugly.

If composer don't want the test suite, this is a composer issue, which doesn't have to be fixed in each package, but in composer.

TomasVotruba:
What exactly is ugly? Contributing to Github repository with help of git is natural. What is your goal actually? 

OT: Composer does not (and should not) restrict downloading of any file. It's up to repository maintainer, what to include.

remicollet:
I'd just like https://github.com/nikic/PHP-Parser/archive/v1.1.0.tar.gz to be usable for downstream distribution.

But close this one if you don't care.

Composer just introduce very bad practices instead of very old good practices (having doc + test in distribution / official tarball).

remicollet:
Notice if you don't distribute "test", no need to distribute phpunit.xml

Ocramius:
Please see doctrine/doctrine2#543 as a reference: this is a useless repetition of the same question.

remicollet:
Sorry, but I don't think this is useless. I still disagree and still think "test" and "doc" should be included.

Else remove everything not needed by composer. Why do you keep README, LICENSE, .*.md ?

Ocramius:
LICENSE *MUST* be distributed, as for README, I actually agree that it's not necessary.

nikic:
Personally I agree with @remicollet - it's not a biggie for this project due to low commit activity, but I find it rather annoying to have to clone huge git repos just to get to the test files.

Comparing the sizes of the 1.0.2 and 1.1.0 tar.gz archives, size went down from 110KB to 74KB. Is that really worth causing extra hassle?

nikic:
I've reverted the `.gitattribute` changes in https://github.com/nikic/PHP-Parser/commit/c96636d19231fd0110021060fb258169b4eacdb3.

remicollet:
Thanks.

DavidPrevot:
:+1:

[...]

PR #303, opened 2016-09-17
Exclude development resources using .gitattributes

Sobak:
More details here: https://www.reddit.com/r/PHP/comments/2jzp6k/i_dont_need_your_tests_in_my_production/

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,12 @@
+/doc             export-ignore
+/grammar         export-ignore
+/tests           export-ignore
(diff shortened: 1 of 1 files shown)


nikic:
See #178. The gitattributes affect more than just composer. In particular downstream distributors do use the tests contained in the tarball.

Sobak:
Ach, I supposed that it has been already proposed but couldn't find related issue/PR. Thanks for the explanation, though.

PR #323, opened 2016-12-04
Add NullableType to types of properties/args that offer it

muglug:
Allows this & similar bugs to be caught: https://github.com/vimeo/psalm/commit/983c6fd6c4e5fbc1b8f18bb496f2a5183ef44614

### DIFF
--- a/lib/PhpParser/BuilderAbstract.php
+++ b/lib/PhpParser/BuilderAbstract.php
@@ -62,7 +62,7 @@ abstract class BuilderAbstract implements Builder {
      *
      * @param Name|string|NullableType $type The type to normalize
      *
-     * @return Name|string The normalized type
+     * @return Name|string|NullableType The normalized type
      */
     protected function normalizeType($type) {
         if (!is_string($type)) {
--- a/lib/PhpParser/BuilderAbstract.php
+++ b/lib/PhpParser/BuilderAbstract.php
@@ -60,7 +60,7 @@ abstract class BuilderAbstract implements Builder {
      * In particular, builtin types are left as strings, custom types become Names and nullables
      * are wrapped in NullableType nodes.
      *
-     * @param Name|string $type The type to normalize
+     * @param Name|string|NullableType $type The type to normalize
      *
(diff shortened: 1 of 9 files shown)


lib/PhpParser/BuilderAbstract.php
```diff
@@ -60,7 +60,7 @@ protected function normalizeName($name) {
      * In particular, builtin types are left as strings, custom types become Names and nullables
      * are wrapped in NullableType nodes.
      *
-     * @param Name|string $type The type to normalize
+     * @param Name|string|NullableType $type The type to normalize
```
nikic: Should be in the `@return` as well

nikic: A few more builder classes are also affected, e.g. FunctionLike::setReturnType and Param::setTypeHint

muglug: Thanks! Added those (and a few others)

nikic:
Thanks!

Seeing this type declaration, I'm thinking that PHP-Parser 4 should have some kind of common superclass `Type` with `BuiltinType`, `ClassLikeType` and `NullableType` as children (or similar).

muglug:
Thanks! Right now the class hierarchy is based on structure - all types in docblocks/typehints are handled as `Union` types, even if they only contain a single `Atomic` type. `Atomic` types can be simple, `Generic` or `ObjectLike`. This means support for nullable types is pretty simple - just checking whether a `Union` contains an `Atomic` type with the value `'null'` - a [simple `isset` check](https://github.com/vimeo/psalm/blob/ff409a7f012fb4c8516006c3176d97196060d56b/src/Psalm/Type/Union.php#L135).

Adding support for `NullableType` as a child class of `Union` would mean a non-trivial update to the logic when doing a null check (where a Union type is cloned, then stripped of its null value) and I'm not sure the benefits outweigh.

But if/when do add `NullableType`, I might also add `FalsableType` (bc PHP stdlib loves those possibly`false` return values), and many libraries follow suit.

PR #324, opened 2016-12-05
NameResolver - resolve NullableType in FunctionLike signature

ondrejmirtes:


lib/PhpParser/NodeVisitor/NameResolver.php
```diff
@@ -163,10 +163,14 @@ private function resolveSignature($node) {
         foreach ($node->params as $param) {
             if ($param->type instanceof Name) {
                 $param->type = $this->resolveClassName($param->type);
+            } elseif ($param->type instanceof Node\NullableType && $param->type->type instanceof Name) {
+                $param->type->type = $this->resolveClassName($param->type->type);
```
nikic: It would be better to handle this as a top-level case in the enterNode method (rather than handling the specific places where it can occur, which are likely to grow)

ondrejmirtes:
I tried a different approach, see the diff. Another place that could have NullableType can just call the wrapName function. I really don't know how to solve this by modifying the enterNode method since the FunctionLike signature is quite nested in the tree.

I'd like if you helped me with this since you have a better idea how to solve this :) Thanks!

The original approach is here https://github.com/ondrejmirtes/PHP-Parser/commit/a25a93a7b7cf85560d44714b2bf25fff5deae793 if you consider it a better starting point.

nikic:
Sorry for being unclear. What I had in mind is just another `elseif` case in `enterNode`: https://github.com/nikic/PHP-Parser/commit/a485ecd7ba007e4d3381471fc46c4721766bcbcc This will handle `NullableType` wherever it may occur.

Also fixed another nullable type issue: https://github.com/nikic/PHP-Parser/commit/5219f7571913f8063de224b231df19ac892645f3

Thanks!

nikic:
New release tagged: https://github.com/nikic/PHP-Parser/releases/tag/v3.0.2

ondrejmirtes:
Great, thank you!

PR #328, opened 2016-12-11
Tweaks to docblocks

muglug:
This fixes a few basic issues flagged by [Psalm](http://getpsalm.org), the static analysis tool I've been working on. Most are small tweaks to docblocks that, I think, improve fellow humans' understanding of the code.

I generated the errors using [this config](https://github.com/muglug/PHP-Parser/commit/8feaef6b56955f8af9a36ebdbd0445f0b8077399#diff-1cb0d3a11875d47ea53a918a026eec8c), not included in the PR (because I don't want to impose any particular tool).

### DIFF
--- a/lib/PhpParser/Node/Expr/Closure.php
+++ b/lib/PhpParser/Node/Expr/Closure.php
@@ -18,7 +18,7 @@ class Closure extends Expr implements FunctionLike
     public $uses;
     /** @var null|string|Node\Name|Node\NullableType Return type */
     public $returnType;
-    /** @var Node[] Statements */
+    /** @var (Node\Stmt|Node\Expr)[] Statements */
     public $stmts;
 
     /**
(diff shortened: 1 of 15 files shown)


lib/PhpParser/Node/Expr/Closure.php
```diff
@@ -18,7 +18,7 @@ class Closure extends Expr implements FunctionLike
     public $uses;
     /** @var null|string|Node\Name|Node\NullableType Return type */
     public $returnType;
-    /** @var Node[] Statements */
+    /** @var Node\Stmt[] Statements */
```
muglug: oh ok – fixed!

muglug:
I've removed the typehints that were necessary for the static analysis tool to understand the code, having improved the tool to compensate.

nikic:
Sorry for the delay, now merged as https://github.com/nikic/PHP-Parser/commit/7b441d2142d34fcf499d155caea53f322e737e43. The (Expr|Stmt)[] is no longer necessary in master because expression statements are represented using a Stmt\Expression node. Some typehints in master are outdated because they don't reflect Identifier nodes ... I'm holding that off until its clear in which form those will stay.

PR #331, opened 2016-12-17
Save original name in NameResolver

### LINKED ISSUE #330: NameResolver should save original name
When doing a find-all-references for a namespace, I only want to return nodes that actually explicitly reference the namespace by its name - that is either a qualified class name or a `use` for the namespace.
Currently this is not possible when using the `NameResolver`, because it will rewrite all names to fully qualified ones and the original name is lost. It would be nice to have an `originalName` attribute added.

felixfbecker:
Closes #330

### DIFF
--- a/lib/PhpParser/NodeVisitor/NameResolver.php
+++ b/lib/PhpParser/NodeVisitor/NameResolver.php
@@ -173,7 +173,11 @@ class NameResolver extends NodeVisitorAbstract
         }
     }
 
-    protected function resolveClassName(Name $name) {
+    protected function resolveClassName(Name $namespacedName) {
+        $name = clone $namespacedName;
+        // Save the original name
+        $name->setAttribute('namespacedName', $namespacedName);
+
         // don't resolve special class names
         if (in_array(strtolower($name->toString()), array('self', 'parent', 'static'))) {
             if (!$name->isUnqualified()) {
@@ -182,7 +186,6 @@ class NameResolver extends NodeVisitorAbstract
                     $name->getAttributes()
                 ));
             }
-
             return $name;
         }
 
--- a/lib/PhpParser/NodeVisitor/NameResolver.php
+++ b/lib/PhpParser/NodeVisitor/NameResolver.php
@@ -205,7 +205,11 @@ class NameResolver extends NodeVisitorAbstract
         return FullyQualified::concat($this->namespace, $name, $name->getAttributes());
     }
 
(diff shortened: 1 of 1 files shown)


lib/PhpParser/NodeVisitor/NameResolver.php
```diff
@@ -202,7 +205,11 @@ protected function resolveClassName(Name $name) {
         return FullyQualified::concat($this->namespace, $name, $name->getAttributes());
     }
 
-    protected function resolveOtherName(Name $name, $type) {
+    protected function resolveOtherName(Name $originalName, $type) {
+        $name = clone $originalName;
+        // Save the original name
+        $name->setAttribute('originalName', $originalName);
```
nikic: For the case of namespace-fallback, this will do something like:

    $name->originalName = clone $originalName;
    $name->namespacedName = FullyQualified::concat($namespace, $name);
    return $name;

In this case $name is still the original name with an extra attribute, so originalName is redundant. However, I guess it's still good to do this for consistency.

felixfbecker: Sorry, I don't get your code snippet?

nikic: I'm referring to the case when https://github.com/felixfbecker/PHP-Parser/blob/1b5969cc36ae1ef98f6acfa0a78c2cc24dd5739c/lib/PhpParser/NodeVisitor/NameResolver.php#L241-L245 is hit.

nikic:
An alternative approach would be to add a mode where nodes are not modified in-place, and instead a `resolvedName` attribute is added. I vaguely remember someone requesting this in the past, because they needed resolved names, but didn't want to change the AST. Doing it this way would also cover your use-case. However, for most purposes this is probably a lot less ergonomic (as you are more often interested in resolved names than original names), as such your approach looks reasonable to me.

felixfbecker:
The `resolvedName` approach would have the disadvantage that pretty-printing the AST would not result in fully-qualified names, which I do... And it would be a big breaking change, because everyone would have to adapt their code. As such, I think it is better to just add the `originalName`

nikic:
Merged via https://github.com/nikic/PHP-Parser/commit/3e8c8d248dcaa837bdb91b82e7883fd9cc055aa5, for now behind a `preserveOriginalNames` option.

Thanks!

felixfbecker:
Thanks a lot!

PR #342, opened 2017-01-23
Add return types to methods where they can be inferred

muglug:
Running my static analysis tool with `--update-docblocks` produces this diff (which I thought may be useful in light of [these changes](https://github.com/nikic/PHP-Parser/commit/670ab2f1788d3548f5370b39c7c29febf100db5a)).

### DIFF
--- a/lib/PhpParser/Comment.php
+++ b/lib/PhpParser/Comment.php
@@ -114,6 +114,9 @@ class Comment implements \JsonSerializable
         return $text;
     }
 
+    /**
+     * @return float
+     */
     private function getShortestWhitespacePrefixLen($str) {
         $lines = explode("\n", $str);
         $shortestPrefixLen = INF;
@@ -127,6 +130,10 @@ class Comment implements \JsonSerializable
         return $shortestPrefixLen;
     }
 
+    /**
+     * @return       array
+     * @psalm-return array{nodeType:string, text:mixed, line:mixed, filePos:mixed}
+     */
     public function jsonSerialize() {
         // Technically not a node, but we make it look like one anyway
         $type = $this instanceof Comment\Doc ? 'Comment_Doc' : 'Comment';
--- a/lib/PhpParser/Parser/Multiple.php
+++ b/lib/PhpParser/Parser/Multiple.php
@@ -23,6 +23,10 @@ class Multiple implements Parser {
         $this->parsers = $parsers;
     }
 
+    /**
+     * @return       null|Error|\PhpParser\Node\Stmt[]
+     * @psalm-return null|Error|array<mixed, \PhpParser\Node\Stmt>
+     */
     public function parse($code, ErrorHandler $errorHandler = null) {
         if (null === $errorHandler) {
             $errorHandler = new ErrorHandler\Throwing;
@@ -43,6 +47,10 @@ class Multiple implements Parser {
         throw $firstError;
     }
 
+    /**
+     * @return       (null|Error|\PhpParser\Node\Stmt[])[]
+     * @psalm-return array<int, null|Error|array<mixed, \PhpParser\Node\Stmt>>
+     */
     private function tryParse(Parser $parser, ErrorHandler $errorHandler, $code) {
         $stmts = null;
         $error = null;
--- a/lib/PhpParser/Parser/Multiple.php
+++ b/lib/PhpParser/Parser/Multiple.php
@@ -24,8 +24,8 @@ class Multiple implements Parser {
     }
 
     /**
-     * @return       null|Error|\PhpParser\Node\Stmt[]
-     * @psalm-return null|Error|array<mixed, \PhpParser\Node\Stmt>
+     * @return       \PhpParser\Node\Stmt[]
+     * @psalm-return array<mixed, \PhpParser\Node\Stmt>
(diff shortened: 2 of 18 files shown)


nikic:
Is it possible to disable generation of `@return void` annotations? I'm not using those anywhere, as I consider the absence of `@return` in a doc-block to indicate that there is no meaningful return value.

muglug:
Sure! I've [added a config](https://github.com/vimeo/psalm/commit/8ec63493d82a036c8c156ff4d1dd4cc812c8b210) to control whether `void` return types are added, and I've updated this PR's commit accordingly.

lib/PhpParser/Comment.php
```diff
@@ -114,6 +114,9 @@ public function getReformattedText() {
         return $text;
     }
 
+    /**
+     * @return float
```
nikic: Shouldn't this type infer to `int|float`, as `strlen()` returns `int`?

lib/PhpParser/Parser/Multiple.php
```diff
@@ -43,6 +47,10 @@ public function parse($code, ErrorHandler $errorHandler = null) {
         throw $firstError;
     }
 
+    /**
+     * @return       (null|Error|\PhpParser\Node\Stmt[])[]
+     * @psalm-return array<int, null|Error|array<mixed, \PhpParser\Node\Stmt>>
```
nikic: Seen above that Psalm also supports something like `@psalm-return array{nodeType:string, text:mixed, line:mixed, filePos:mixed}`. If this here were `@psalm-return array{0:null|Node\Stmt[], 1:null|Error}` then the `list()` unpacking in the method above could see that `Error` is not a possible return type.

nikic:
Overall, nice results. There's a few cases where the return type is too broad (e.g. `NameResolver::getResolvedClassName()` can't return null), but inferring that would be non-trivial (something like CPA could determine that null is only returned if both arguments are null and in this case at most one of them is).

muglug:
Thanks! To that last point, how about throwing an exception if both `$name1` and `$name2` are `null` in `PhpParser\Node\Name::concat`? It seems like that can never happen, at least in PhpParser's own usage of that method. Doing so would erase the null value from that return type, and solve that particular issue.

lib/PhpParser/Parser/Multiple.php
muglug: Unfortunately `0` and `1` as keys imply that they are strings, so argument unpacking wouldn't work there. A solution is either to add string array offsets (and change what the method returns, and how it's used above) or just strip out the return types for now.

Obviously PHP 7.1's array key destructuring would simplify *that* diff, but happy that PhpParser is PHP 5-compatible.

nikic: PHP does not distinguish between integer keys and integral string keys. So here all of `[$a, $b]`, `[0 => $a, 1 => $b]` and `["0" => $a, "1" => $b]` are equivalent (both in terms of array creation and unpacking). I'm fine with just stripping the return type, just wondering why Psalm makes this distinction, i.e., why are "shapes" only created for arrays with string keys, rather than also arrays with fixed-structure integral keys? Is this because there might be false positives where such a return value is intended as a normal array (which just happens to have the same non-uniform types on all paths) rather than a shape?

muglug:
Yup, there's no reason why Psalm shouldn't be able to understand that formulation for `ObjectLike` types when processing `ArrayDimFetch` and when destructuring lists. I will ticket and add soonish. 

Generating that type is considerably less trivial (as there are a bunch of different desired outcomes depending on the different return statements in a function), but I'll ticket that as well.

lib/PhpParser/ParserAbstract.php
```diff
@@ -602,6 +615,9 @@ protected function fixupStartAttributes(Node $to, Node $from) {
         }
     }
 
+    /**
+     * @return string
```
nikic: Noticed while merging: This return type is off. It should be inferred as `string|Name|Node\Identifier`, as there are `return $name`s and a `return new Node\Identifier` via ternary.

nikic:
Merged via https://github.com/nikic/PHP-Parser/commit/e3b87f40aa6d2ca65ed40ee26566299412985141, with more updates in https://github.com/nikic/PHP-Parser/commit/bfea338d36721ba91db849d0596d07a1d585bcc0, where I manually improved some of the types and added descriptions + param annotations for everything.

I did not merge the changes to the Standard pretty printer, as there basically all methods return strings and I think adding explicit doc comments there is more noisy than helpful.

Thanks!

lib/PhpParser/ParserAbstract.php
muglug: Ah bugger. Sorry. It's because both classes have a `__toString` method, and Psalm gets confused when merging the types. Fixing...

lib/PhpParser/Comment.php
muglug: There was existing logic to support `int`s being cast to `float`s in function calls, and it (falsely) trickled into the type simplification logic – fixed here: https://github.com/vimeo/psalm/commit/5ec2a9742b3e39c6320509cb9f2e565da1505ac3

PR #348, opened 2017-02-02
Tighten method parameters

muglug:
This PR makes a few minor tweaks to docblocks, and adds a little protection for types.

### DIFF
--- a/lib/PhpParser/Lexer.php
+++ b/lib/PhpParser/Lexer.php
@@ -166,7 +166,7 @@ class Lexer
                 $nextFilePos = strpos($this->code, $tokenValue, $filePos);
                 $this->handleInvalidCharacterRange(
                     $filePos, $nextFilePos, $line, $errorHandler);
-                $filePos = $nextFilePos;
+                $filePos = (int)$nextFilePos;
             }
 
             $filePos += $tokenLen;
--- a/lib/PhpParser/NodeTraverser.php
+++ b/lib/PhpParser/NodeTraverser.php
@@ -129,6 +129,10 @@ class NodeTraverser implements NodeTraverserInterface
                     }
                 }
 
+                if (is_int($subNode)) {
+                    throw new \UnexpectedValueException('$subNode should not be an int');
+                }
+
                 if ($traverseChildren) {
                     $subNode = $this->traverseNode($subNode);
                     if ($this->stopTraversal) {
@@ -165,9 +169,11 @@ class NodeTraverser implements NodeTraverserInterface
     /**
      * Recursively traverse array (usually of nodes).
      *
-     * @param array $nodes Array to traverse
+     * @param (Node|Node[])[] $nodes Array to traverse
      *
-     * @return array Result of traversal (may be original array or changed one)
+     * @return Node[] Result of traversal (may be original array or changed one)
+     * 
+     * @psalm-suppress MoreSpecificReturnType
      */
     protected function traverseArray(array $nodes) {
         $doNodes = array();
@@ -191,6 +197,10 @@ class NodeTraverser implements NodeTraverserInterface
                             );
                         }
                     }
+                }
+
+                if (is_int($node)) {
+                    throw new \UnexpectedValueException('$node should not be an int');
                 }
 
                 if ($traverseChildren) {
--- a/lib/PhpParser/NodeTraverser.php
+++ b/lib/PhpParser/NodeTraverser.php
@@ -102,6 +102,7 @@ class NodeTraverser implements NodeTraverserInterface
      */
     protected function traverseNode(Node $node) {
         foreach ($node->getSubNodeNames() as $name) {
+            /** @var Node[] */
             $subNode =& $node->$name;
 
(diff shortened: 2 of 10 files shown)


lib/PhpParser/PrettyPrinter/Standard.php
```diff
@@ -530,6 +530,10 @@ protected function pExpr_ConstFetch(Expr\ConstFetch $node) {
     }
 
     protected function pExpr_ClassConstFetch(Expr\ClassConstFetch $node) {
+        if (!is_string($node->name)) {
+            throw new \LogicException('$node->name must be a string');
+        }
```
nikic: Fixed this issue in a different way with https://github.com/nikic/PHP-Parser/commit/d287c167bc3cc9cc34e3e4da3a7d9282b14b6e90. As this is a genuine bug landed in 3.x. (The check implemented here is causing the build failure on Travis.)

lib/PhpParser/NodeTraverser.php
```diff
@@ -154,9 +158,11 @@ protected function traverseNode(Node $node) {
     /**
      * Recursively traverse array (usually of nodes).
      *
-     * @param array $nodes Array to traverse
+     * @param (Node|Node[])[] $nodes Array to traverse
```
nikic: This annotation isn't correct, because the array doesn't necessarily contain nodes. For example, the `Name::$parts` subnode is an array of strings.

Also, `Node[][]` is not actually possible here -- I just dropped the code for handling this case and made it throw an exception instead: https://github.com/nikic/PHP-Parser/commit/2b6804aa507354e89f2cd108858fb3d74027e2e3

lib/PhpParser/NodeTraverser.php
```diff
@@ -181,6 +187,10 @@ protected function traverseArray(array $nodes) {
                     }
                 }
 
+                if (is_int($node)) {
+                    throw new \UnexpectedValueException('$node should not be an int');
+                }
```
nikic: I've implemented a stricter version of this check in https://github.com/nikic/PHP-Parser/commit/2b72bae3d9ad3eeef4f013de584cfd28963d7c50. Now all return values from enterNode and leaveNode which are not explicitly handled in some way will throw.

lib/PhpParser/Lexer.php
```diff
@@ -166,7 +166,7 @@ protected function handleErrors(ErrorHandler $errorHandler) {
                 $nextFilePos = strpos($this->code, $tokenValue, $filePos);
                 $this->handleInvalidCharacterRange(
                     $filePos, $nextFilePos, $line, $errorHandler);
-                $filePos = $nextFilePos;
+                $filePos = (int)$nextFilePos;
```
nikic: Why is this cast necessary? Does it exclude a possible `false` return value from `strpos`?

lib/PhpParser/ParserAbstract.php
```diff
@@ -574,6 +574,12 @@ private function getNamespacingStyle(array $stmts) {
         return $style;
     }
 
+    /**
+     * @param  Node\Expr\StaticPropertyFetch|Node\Expr\ArrayDimFetch $prop
+     * @param  Node\Arg[]       $args
+     * @param  string[]         $attributes
```
nikic: This one should be just `array`.

lib/PhpParser/Lexer.php
muglug: Yes – really, it's just telling Psalm not to worry. I could also put a `/** @var int **/ declaration above the `strpos` call, which would prevent the actual code from becoming polluted.

muglug:
I've just touched this up a wee bit. Happy to squash the commits.

lib/PhpParser/NodeVisitor/FirstFindingVisitor.php
```diff
@@ -33,6 +33,8 @@ public function getFoundNode() {
 
     public function beforeTraverse(array $nodes) {
         $this->foundNode = null;
+
+        return null;
```
GrahamCampbell: 👎

muglug: `FirstFindingVisitor::beforeTraverse` inherits the return type `null|Node[]`, and this explicit (and I agree, wholly unnecessary given PHP's internals) `return` statement acknowledges that.

nikic:
Merged via https://github.com/nikic/PHP-Parser/commit/329e90c239267342199dd73a72642c15d81faf05 with two omissions: I've left out the NodeTraverser changes as the `Node[]` annotation is not always correct (see [this comment](https://github.com/nikic/PHP-Parser/pull/348#discussion_r99428031)). I've also left out the [Namespace node](https://github.com/nikic/PHP-Parser/pull/348/files#diff-ae3df4c564fd4d694ac8f7042e672f4fR11) change -- this one is of course correct, but the null value only exists temporarily during parsing, the end user never sees it.

muglug:
Thanks!

Issue #351, opened 2017-02-08
Comments inside arrays are dropped by Standard pretty printer


jhodgdon-drp:
If you have some code like
```
array(
  'foo' => 'bar',
  // Comment here.
  'baz' => 7,
);
```
(or similar with [] array shorthand), and you pretty print it, it comes out as one line:
`array('foo' => 'bar', 'baz' => 7);`

So, the comment gets dropped.

I'm making a Pretty Printer class for a project, and the solution I came up with was to output all arrays multi-line instead of in one line, and then each Expr_ArrayItem can print out its comment (which is there in the parser output statements, but dropped currently by the Standard pretty printer).

jhodgdon-drp:
By the way THANK YOU for this project! We had our own parser and pretty printer in our project, and using this library has made our code much cleaner, more understandable, and more maintainable. I'm just going to file a few issues about the Pretty Printer class that I came across... they aren't holding me up though since we needed our own pretty printer class anyway. Thanks!

jhodgdon-drp:
I'll make a pull request with the code I used to accomplish this. Fairly straightforward...

nikic:
This issue and #366 are now resolved by https://github.com/nikic/PHP-Parser/commit/c3cbf079466fad893e576d072c1f9daebfe04f95, which is a conservative version of #355, printing on multiple lines only if it is necessary to preserve comments.

Printing all arrays as multiline is something we could have behind an option (or maybe a callback so one can for example decide based on number of elements?)

PR #355, opened 2017-02-08
Print arrays multi-line with comments

jhodgdon-drp:
See issue #351


nikic:
Some of the pretty printer tests that check the format for arrays are failing. Looking at some of the diffs, we have things like

```
-[, [[$a]], $b] = $bar;
+[
+    ,
+    [
+        [
+            $a,
+        ],
+    ],
+    $b,
+] = $bar;
```

which makes me think that using multiline format for *all* arrays is probably not appropriate.

jhodgdon-drp:
That may be. I would argue that all multi-line is much better than all in-line, and that particular sample is not code that I see a lot in PHP... certainly discarding comments that are in arrays that are parsed out is also bad.

I guess maybe the array needs some kind of flag to say whether it should be one-line or multi-line, if you want to support both?

nikic:
Closing this based on https://github.com/nikic/PHP-Parser/issues/351#issuecomment-292801293. I went with a conservative variant for now, but of course having an option to print all arrays on multiple lines would be fine.

Issue #338, opened 2017-01-16
skipped installation of bin bin/php-parse for package nikic/php-parser: file not found in package


RobbiNespu:
Why this package are not found? 
```
$ php composer.phar update
Loading composer repositories with package information
Updating dependencies (including require-dev)
Nothing to install or update
Generating autoload files
> Cake\Composer\Installer\PluginInstaller::postAutoloadDump
    Skipped installation of bin bin/php-parse for package nikic/php-parser: file not found in package
```

Here my composer.json
```
{
    "name": "cakephp/app",
    "description": "CakePHP skeleton app",
    "homepage": "http://cakephp.org",
    "type": "project",
    "license": "MIT",
    "require": {
        "php": ">=5.5.9",
        "cakephp/cakephp": "3.3.*",
        "mobiledetect/mobiledetectlib": "2.*",
        "cakephp/migrations": "~1.0",
        "cakephp/plugin-installer": "*",
        "karyamedia/ipay88": "dev-master",
        "geoip2/geoip2": "~2.0",
        "nikic/php-parser": "^3.0"
    },
    "require-dev": {
        "psy/psysh": "@stable",
        "cakephp/debug_kit": "~3.2 !=3.5.1",
        "cakephp/bake": "~1.1"
    },
    "suggest": {
        "markstory/asset_compress": "An asset compression plugin which provides file concatenation and a flexible filter system for preprocessing and minification.",
        "phpunit/phpunit": "Allows automated tests to be run without system-wide install.",
        "cakephp/cakephp-codesniffer": "Allows to check the code against the coding standards used in CakePHP."
    },
    "autoload": {
        "psr-4": {
            "App\\": "src"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "App\\Test\\": "tests",
            "Cake\\Test\\": "./vendor/cakephp/cakephp/tests"
        }
    },
    "scripts": {
        "post-install-cmd": "App\\Console\\Installer::postInstall",
        "post-create-project-cmd": "App\\Console\\Installer::postInstall",
        "post-autoload-dump": "Cake\\Composer\\Installer\\PluginInstaller::postAutoloadDump",
        "check": [
            "@test",
            "@cs-check"
        ],
        "cs-check": "vendor/bin/phpcs --colors -p --standard=vendor/cakephp/cakephp-codesniffer/CakePHP ./src ./tests",
[...]

PR #358, opened 2017-02-15
Fix: fix issue #338

shashankkrjain:
This fixes issue #338. The path of the executable file is wrong, which stops composer install and update.

Executable should be in the root folder and only its name should be given in composer. This prevents bad path names across platforms (like, in Windows).

See following repositories for example: 
https://github.com/sebastianbergmann/phpunit
https://github.com/laravel/installer


nikic:
Do you have a reference regarding the restriction to the top-level directory? The [examples in the composer docs](https://getcomposer.org/doc/articles/vendor-binaries.md) consistently use scripts in the `bin/` directory, so I have a hard time believing that this is not supported.

shashankkrjain:
Sorry for this. There was a weird error I was facing, and I believed this fixed it. However, it didn't and I had to delete vendor folder and reinstall everything to make it work. All plugins with executable were giving errors so this was not particularly this plugin.

PR #361, opened 2017-03-02
Removed double semicolons

pjona:


nikic:
These files are generated, so any changes to them will be discarded on the next run of `grammar/rebuildParsers.php`. As this code isn't intended for human consumption, it probably doesn't make a lot of sense to fix the coding style anyway...

PR #363, opened 2017-03-05
Run on Trusty Beta Container

photodude:
Runs the tests on the newer Trusty containers since this repo is not affected by the mysql issues in the trusty container. 

Added benefit, This provides the current HHVM version (3.18.1 as of this PR) and will track with each release (i.e. will be 3.19 when 3.19 is released). ... **you are currently on an EOL versions of HHVM since precise was dropped back at hhvm 3.6**

If testing against HHVM LST versions is desired follow this guide. https://docs.travis-ci.com/user/languages/php#HHVM-versions

NOTE: If testing against HHVM in php 7 mode is desired: HHVM php 7 mode currently has an issue that messes up composer under php 7 mode, see facebook/hhvm#7198 once that BUG is fixed HHVM PHP 7 mode testing could be added

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: php
-
+dist: trusty
 sudo: false
 
 cache:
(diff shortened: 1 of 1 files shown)


nikic:
Thanks! Unfortunately it doesn't look like binary string lexing was fixed in the meantime, so we still have test failures on HHVM :(

PR #368, opened 2017-03-21
Make the Error class used explicit.

TysonAndre:
Saw the below error in php 7.1 (The base class didn't have a `use`, but the subclass did). Not fully familiar with the rules for checking if unqualified names were compatible, but this fixes it.

Seen in 3.0.5

```
Fatal error: Uncaught Error: Declaration of PhpParser\ErrorHandler\Throwing::handleError() must be compatible with that of PhpParser\ErrorHandler::handleError() in .../vendor/nikic/php-parser/lib/PhpParser/ErrorHandler/
Throwing.php:13
Stack trace:
\#0 .../vendor/nikic/php-parser/lib/PhpParser/Autoloader.php(36): include()
\#1 ..../vendor/nikic/php-parser/lib/PhpParser/ParserAbstract.php(144): PhpParser\Autoloader::autoload()
...
```

### DIFF
--- a/lib/PhpParser/ErrorHandler.php
+++ b/lib/PhpParser/ErrorHandler.php
@@ -2,6 +2,8 @@
 
 namespace PhpParser;
 
+use PhpParser\Error;
+
 interface ErrorHandler
 {
     /**


nikic:
Unqualified class names are relative to the current namespace, so `Error` in this context already refers to `PhpParser\Error`.

TysonAndre:
Sorry. The codebase I used was for php7.1, but I encountered this error trying to run this in hhvm nightly

GrahamCampbell:
Doesn't this use now introduce a conflict with itself, if the error class is already autoloaded?

GrahamCampbell:
This bug should be fixed in HHVM, and no change should be made here IMO.

GrahamCampbell:
Actually, it doesn't look like PHP complains if something of the same name gets imported again, so this change doesn't break anything, but it's still not needed IMO.

TysonAndre:
Running without `hhvm.php7.all=1` is one workaround for that.

Also, this PR is incomplete - I'd have to change the places where `new Error` was called (When the intent was `PhpParser\Error`). E.g. ParserAbstract.php

fredemmott:
Should be fixed in the next HHVM release, https://github.com/facebook/hhvm/commit/29fc86f9ddefc918549c201ebe5ee8a978efde44

PR #372, opened 2017-04-07
Removes unneeded code from JSON encoding example.

Potherca:
Unless I am mistaken the NodeDumper is _not_ needed when outputting to JSOn, so the following line  can be removed from the docs:
```php
$nodeDumper = new PhpParser\NodeDumper;
```


nikic:
Yeah, copy&paste mistake. Thanks!

PR #375, opened 2017-04-15
Move token byref vars into object

muglug:
This removes the only byref function params in the lexer. This doesn't fix any bugs – just makes the code a little easier to reason about.

Feel free not to merge if you think there are performance implications. But I didn't see a negative impact in my own testing, FWIW.


nikic:
Ah, for the record, the reason why I didn't merge this is that during some preliminary performance testing this caused a minor regression (iirc something like 2% end to end). I was thinking it could be improved by not creating a new object every time, but never got around to testing it.

muglug:
yeah, no big deal – architecture wasn't ideal

muglug:
Out of interest, how are were you able to discover such a small regression with significance?

PR #376, opened 2017-04-16
exposing variadic & attributes args for param builder

SignpostMarv:
I needed to use some variadic params in code generation, so I've added methods & props to expose the remaining arguments on Node\Param

### DIFF
--- a/lib/PhpParser/Builder/Param.php
+++ b/lib/PhpParser/Builder/Param.php
@@ -16,6 +16,8 @@ class Param extends PhpParser\BuilderAbstract
 
     protected $byRef = false;
 
+    protected $variadic = false;
+
     /**
      * Creates a parameter builder.
      *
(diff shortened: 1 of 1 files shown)


nikic:
I'm fine with makeVariadic(), but what's the motivation for setAdditionalAttributes()? If we add something like this, it probably should be added for all builders.

SignpostMarv:
it was a case of "if I'm adding this, why not add this other thing?"- if we're not particularly fussed about exposing additionalAttributes, I can force-push it without the additional attributes method?

nikic:
Yeah, I'd prefer that.

SignpostMarv:
done :)

nikic:
Merged as https://github.com/nikic/PHP-Parser/commit/c877c1a64ff7c3dcdcd78ceed292fc4557ed6fd2 into 3.x. Thanks!

Issue #378, opened 2017-04-21
Proposal: add to ClassMethod checking on magic method


DragorWW:
Proposal: 
- add a method for checking if this method - [magic method](http://php.net/manual/en/language.oop5.magic.php)

**Node\Stmt\ClassMethod**
- isMagic(): bool

_Thanks for this project, @nikic. Do you mind, if I will suggesting improvements to the api?_

nikic:
I'm fine with adding a toMagic() method. Do you plan to PR this?

DragorWW:
@nikic yes, i will can create PR for this functional.

PR #379, opened 2017-04-22
ClassMethod: add checking on magic method

DragorWW:
code for #378 issue

changes:
- ClassMethod: add method isMagic
- ClassMethodTest: add test for isMagic

_All is well, or need to change?_

### DIFF
--- a/lib/PhpParser/Node/Stmt/ClassMethod.php
+++ b/lib/PhpParser/Node/Stmt/ClassMethod.php
@@ -7,24 +7,6 @@ use PhpParser\Node\FunctionLike;
 
 class ClassMethod extends Node\Stmt implements FunctionLike
 {
-    const MAGIC_NAMES = array(
-        '__construct',
-        '__destruct',
-        '__call',
-        '__callstatic',
-        '__get',
-        '__set',
-        '__isset',
-        '__unset',
-        '__sleep',
-        '__wakeup',
-        '__tostring',
-        '__set_state',
-        '__clone',
-        '__invoke',
-        '__debuginfo',
-    );
-
     /** @var int Flags */
     public $flags;
     /** @var bool Whether to return by reference */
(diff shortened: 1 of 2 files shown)


lib/PhpParser/Node/Stmt/ClassMethod.php
```diff
@@ -117,4 +135,13 @@ public function isFinal() {
...
         return (bool) ($this->flags & Class_::MODIFIER_STATIC);
     }
+
+    /**
+     * Whether the method is magic.
+     *
+     * @return bool
+     */
+    public function isMagic() {
+        return (bool) in_array($this->name, static::MAGIC_NAMES);
```
nikic: Method names in PHP are case-insensitive, so this should do a `strtolower` call first.

nikic: Additionally, I would suggest to convert `MAGIC_NAMES` into a map (see for example https://github.com/DragorWW/PHP-Parser/blob/00d3ca3ccdeb14480a55af6c47d1ca9bd5e9cde7/lib/PhpParser/Node/Stmt/Class_.php#L28) and then check whether it is a magic method using `isset()`. This is more efficient than using in_array.

DragorWW: @nikic may be `const MAGIC_NAMES` change to `protected static $magicNames`?

nikic: Yeah, looks like the const array currently breaks the build on 5.6 and on HHVM.

DragorWW: @nikic fix it

nikic:
Merged as https://github.com/nikic/PHP-Parser/commit/9dc93aafaab90eb885a59412237b323047c1819a. Thanks!

PR #381, opened 2017-04-26
Bump PHPUnit version

ekinhbayar:
Happened to do this when I was playing with something else, PR'd in case you may want to bump the version.

Also, there's a PHPUnit config option (beStrictAboutTestsThatDoNotTestAnything) that you could set to false to avoid the risky tests flood but I doubt that'd be the correct way to do so.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "ext-tokenizer": "*"
     },
     "require-dev": {
-        "phpunit/phpunit": "~5.0"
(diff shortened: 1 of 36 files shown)


nikic:
Thanks! Also added the mentioned option in https://github.com/nikic/PHP-Parser/commit/e1ab92275cefe13c8e41c0ca42679258ebecbce8.

PR #390, opened 2017-05-28
Fix example code to work

xKerman:
Example code in [Example: Converting namespaced code to pseudo namespaces](https://github.com/nikic/PHP-Parser/blob/master/doc/2_Usage_of_basic_components.markdown#example-converting-namespaced-code-to-pseudo-namespaces) does not convert namespaces to pseudo namespaces,
since `PhpParser\Node\Name::toString()` does not accept namespace separator from this commit: a910f6a247b9041401558a0a4335773103848c89

In this pull request, to convert namespaces, I replaced pseudo namespace generation code.


nikic:
Thanks!

PR #391, opened 2017-05-31
Fix Lexer errorHandling when there is no tokens

romainneutron:
I encounter an issue when no tokens exist. An error `Undefined offset -1` is triggered.

### DIFF
--- a/lib/PhpParser/Lexer.php
+++ b/lib/PhpParser/Lexer.php
@@ -182,15 +182,17 @@ class Lexer
             return;
         }
 
-        // Check for unterminated comment
-        $lastToken = $this->tokens[count($this->tokens) - 1];
-        if ($this->isUnterminatedComment($lastToken)) {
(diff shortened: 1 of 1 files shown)


nikic:
Under which circumstances do you see the notice? Is it an empty file on HHVM perchance? The change makes sense in any case, just wondering how to reproduce.

romainneutron:
I don't reproduce the issue anymore. However, I remember it was using symfony php 7.0 polyfill with PHP 5.5. This [part of the code of the polyfill](https://github.com/symfony/polyfill-php70/blob/master/Php70.php#L59) triggered an error, leading to `Lexer::errorMayHaveOccurred` returning `true`.

romainneutron:
Does it help you?

nikic:
Merged as https://github.com/nikic/PHP-Parser/commit/c28b8556f512cbcbcdcefa48abc8020751f997cd into 3.x. Thanks :)

PR #394, opened 2017-06-30
Fixed incorrect class inheritance in code generation example (docs)

uniconstructor:
Generated example was incorrect - you cannot define import alias:
use Some\Other\Thingy as SomeClass;

and then define a class with the same name:
abstract class SomeClass ...

Class names changed to avoid conflict between import alias and abstract class name.


TomasVotruba:
Yep, this makes sense.

PR #405, opened 2017-08-07
Docs: highlight interface + PHP 7 return types

TomasVotruba:
To make it more natural and readable

### DIFF
--- a/doc/component/Lexer.markdown
+++ b/doc/component/Lexer.markdown
@@ -96,10 +96,10 @@ Lexer extension
 A lexer has to define the following public interface:
 
 ```php
-startLexing(string $code, ErrorHandler $errorHandler = null): void;
-getTokens(): array;
-handleHaltCompiler(): string;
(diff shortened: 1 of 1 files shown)


doc/component/Lexer.markdown
```diff
@@ -95,10 +95,12 @@ Lexer extension
...
 
-    void startLexing(string $code, ErrorHandler $errorHandler = null);
-    array getTokens();
-    string handleHaltCompiler();
-    int getNextToken(string &$value = null, array &$startAttributes = null, array &$endAttributes = null);
+```php
+startLexing(string $code, ErrorHandler $errorHandler = null): void;
+getTokens(): array;
+handleHaltCompiler(): string;
+getNextToken(string &$value = null, array &$startAttributes = null, array &$endAttributes = null): int;
```
nikic: With this format I'd suggest adding the `function` prefix.

TomasVotruba: Good point, I've added it

TomasVotruba:
Thank you

PR #407, opened 2017-08-13
travis: add dash to make command format united

TomasVotruba:
Prevents possible error when adding a second line


TomasVotruba:
:+1:

PR #410, opened 2017-08-13
Add strict_types declarations 

TomasVotruba:
I've noticed PHP 7 typehints are already there, so could complete them.

I came across only few `strtolower($node)` => `strtolower((string) $node` issues, all works otherwise.

### DIFF
--- a/bin/php-parse
+++ b/bin/php-parse
@@ -26,9 +26,9 @@ if (empty($files)) {
     showHelp("Must specify at least one file.");
 }
 
-$lexer = new PhpParser\Lexer\Emulative(array('usedAttributes' => array(
+$lexer = new PhpParser\Lexer\Emulative(['usedAttributes' => [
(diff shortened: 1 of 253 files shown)


nikic:
Strict types addition merged as https://github.com/nikic/PHP-Parser/commit/e2e99f269b5b543c9196e1cf56348942165e8faf. I've handled the extra string casts a bit differently in https://github.com/nikic/PHP-Parser/commit/d97cc3d96e661e199244dc993bced393e3cd231d and https://github.com/nikic/PHP-Parser/commit/7b36ca3b6cc1b99210c6699074d6091061e73eea.

TomasVotruba:
Nice, I prefer your solution. Thanks for update :+1:

TomasVotruba:
Shall I send `strict_types` to tests as well or did you skip that for some reason?

PR #416, opened 2017-09-16
Added a missing property declaration for pretty printer

markbook2:


markbook2:
I had not found any occurrences of the PrettyPrinterAbstract::noIndentToken property anywhere except the constructor of the PrettyPrinterAbstract. Can it be removed instead of adding the declaration, proposed here?

nikic:
Yeah, it can be removed. It's a leftover from a recent refactor of indentation handling.

PR #423, opened 2017-10-05
Add pretty printer interface

ircmaxell:
I've customized the printer in the past, and having a simple interface for it could be interesting. The use-case isn't compelling, but it feels like a nice symmetry between the rest of the interfaces and abstract classes.

### DIFF
--- a/lib/PhpParser/PrettyPrinter.php
+++ b/lib/PhpParser/PrettyPrinter.php
@@ -2,7 +2,7 @@
 
 namespace PhpParser;
 
-use Node\Expr;
+use PhpParser\Node\Expr;
 
 interface PrettyPrinter
 {
--- a/lib/PhpParser/PrettyPrinter.php
+++ b/lib/PhpParser/PrettyPrinter.php
@@ -1,0 +1,17 @@
+<?php declare(strict_types=1);
+
+namespace PhpParser;
+
+interface PrettyPrinter
+{
+
+    /**
+     * Convert the AST into PHP code suitable for execution
+     *
+     * @param Node[] $stmts The statements to render
+     * 
+     * @return string Rendered PHP code
+     */
+    public function prettyPrint(array $stmts) : string;
+
+}
--- a/lib/PhpParser/PrettyPrinter.php
+++ b/lib/PhpParser/PrettyPrinter.php
@@ -2,16 +2,60 @@
 
 namespace PhpParser;
 
+use Node\Expr;
+
 interface PrettyPrinter
 {
 
     /**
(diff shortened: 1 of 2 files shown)


nikic:
The public interface of `PrettyPrinterAbstract` has a couple more methods:

```php
interface PrettyPrinter {
    public function prettyPrint(array $stmts) : string;
    public function prettyPrintExpr(Expr $node) : string;
    public function prettyPrintFile(array $stmts) : string;
    public function printFormatPreserving(array $stmts, array $origStmts, array $origTokens) : string;
}
```

The thing is, I think that any implementation of the `PrettyPrinter` interface is going to extend from `PrettyPrinterAbstract` (and quite likely from the `Standard` printer as well -- there are large parts that one likely does not want to adjust).

Would I be mistaken in the assumption that what you're looking for here is not actually an interface, but rather the ability to write `PrettyPrinter` rather than `PrettyPrinterAbstract` in type hints?

ircmaxell:
> Would I be mistaken in the assumption that what you're looking for here is not actually an interface, but rather the ability to write PrettyPrinter rather than PrettyPrinterAbstract in type hints?

Well, yes and no.

Right now there's a single implementation of `PrettyPrinter` (`PrettyPrinter\Standard`). The two classes today (Abstract and Standard) are quite tightly coupled in that most use-cases would extend Standard instead of Abstract just due to the nature of the implementation.

So that raises a significant question: what's the purpose of Abstract in the first place? 

Introducing an interface for the type would allow Abstract to pull in the default implementation of Standard, and let Standard simply an empty extension. Then that opens the door to extending Abstract simply to render exactly what's needed (overriding the Abstract methods as needed).

For example, to implement a pretty printer that strips out comments, you would simply need to extend abstract and override `pComments` to always return `''`. 

Again, not a huge win, but seems that the symmetry with the rest of the design may be worth while.

Pushing a fix adding the other 3 methods + some docs...

lib/PhpParser/PrettyPrinter.php
```diff
@@ -0,0 +1,61 @@
...
+     * @param Node[] $stmts      Modified AST with links to original AST
+     * @param Node[] $origStmts  Original AST with token offset information
+     * @param array  $origTokens Tokens of the original code
+     *
+     * @return string
+     */
+    public function printFormatPreserving(array $stmts, array $origStmts, array $origTokens) : string;
+
+
+}
```
ronfroy: You should maybe add a line break here

nikic:
I've decided to merge this after all, for the next major version. It's only been ... six years? Ooops.

TomasVotruba:
🎉

PR #427, opened 2017-10-18
Easy way to fiddle with the traverser.

synga-nl:
Sometimes I want to quickly develop something without the need of ceating a visitor class. I created this class and I think more people want to use this.

### DIFF
--- a/lib/PhpParser/NodeVisitor/CallableVisitor.php
+++ b/lib/PhpParser/NodeVisitor/CallableVisitor.php
@@ -1,0 +1,95 @@
+<?php
+namespace PhpParser\NodeVisitor;
+
+use PhpParser\Node;
+use PhpParser\NodeVisitorAbstract;
+
+/**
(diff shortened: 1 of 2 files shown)


nikic:
What is the motivation for using this over an anonymous class?

TomasVotruba:
I've just randomly found `NodeFinder` class: 

- https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/NodeFinder.php
- https://github.com/nikic/PHP-Parser/blob/master/test/PhpParser/NodeFinderTest.php

Maybe you could find it useful

nikic:
I'm closing this, as a) you can already use an anonymous class with about the same amount of effort (https://github.com/nikic/PHP-Parser/blob/master/doc/component/Walking_the_AST.markdown#walking-the-ast) and b) the specific case of finding nodes is handled by NodeFinder (https://github.com/nikic/PHP-Parser/blob/master/doc/component/Walking_the_AST.markdown#simple-node-finding).

PR #435, opened 2017-11-01
ClassMethod stmts property can be null

codeliner:
See this comment for details: https://github.com/Roave/BetterReflection/issues/378#issuecomment-341089928


lib/PhpParser/Node/Stmt/ClassMethod.php
```diff
@@ -17,7 +17,7 @@ class ClassMethod extends Node\Stmt implements FunctionLike
     public $params;
     /** @var null|Node\Identifier|Node\Name|Node\NullableType Return type */
     public $returnType;
-    /** @var Node\Stmt[] Statements */
+    /** @var Node\Stmt[]|Null Statements */
```
nikic: Can you please change this to a lowercase `null`, to be consistent with other type declarations?

codeliner:
@nikic of course, done

PR #436, opened 2017-11-02
Add strict_types to tests as well

TomasVotruba:
To be compatible with `/lib`

There was just 1 issue to handle


TomasVotruba:
:+1:

PR #441, opened 2017-11-11
Correctly determine Type of Node when PHP-Parser's namespaces are prefixed

borNfreee:
Hi there,

I'm working on mutation testing framework ([Infection](https://github.com/infection/infection/)) that is distributed as a PHAR. One of its goal is to run target project's test suite against mutated code. Since we use reflection and load project's autoloader, we want to avoid potential conflicts between vendor files of Infection itself and the target project.

To avoid this issue, there is a project called [PHP-Scoper](https://github.com/humbug/php-scoper). What it does is it prefixes all the namespaces of the library (including vendor folder) with some character(s), for example namespace `Infection\Mutator\PublicVisibility` is transformed to `ScoperAbc123\Infection\Mutant\PublicVisibility`. 

But since it also prefixes vendor folder, PHP-Parser's classes are prefixed as well and `NodeAbstract::getType()` after this prefixing works incorrectly.

There is a hardcoded number `15` which means to remove `'PhpParser\Node'` (length=15) substring from the FQCN.

Code:

```php
// PHPParser\Node\Stmt\Declare_ -> Stmt_Declare

return strtr(substr(rtrim(get_class($this), '_'), 15), '\\', '_');
```

What I suggest is a little bit more dynamic solution, to correctly extract class name (type) from the **prefixed** FQCN:

`ScoperAbc123\PHPParser\Node\Stmt\Declare_` -> `Stmt_Declare`

### DIFF
--- a/lib/PhpParser/NodeAbstract.php
+++ b/lib/PhpParser/NodeAbstract.php
@@ -21,7 +21,15 @@ abstract class NodeAbstract implements Node, \JsonSerializable
      * @return string Type of the node
      */
     public function getType() : string {
-        return strtr(substr(rtrim(get_class($this), '_'), 15), '\\', '_');
+        $className = rtrim(get_class($this), '_');
+        return strtr(
+            substr(
+                $className,
+                strpos($className, 'PhpParser\Node') + 15
+            ),
+            '\\',
+            '_'
+        );
     }
 
     /**


nikic:
Merged https://github.com/nikic/PHP-Parser/commit/bac91b426e721cc7183208199c525d6c43ab63c6 into 3.x. For 4.0 I've generated explicit getType() methods instead: https://github.com/nikic/PHP-Parser/commit/1c11626f0a2fda566d9db9f388feb0682093955a

borNfreee:
Great, thank you.

theofidry:
@borNfreee you may also ran into the issue with the Lexer cf. https://github.com/humbug/php-scoper/blob/master/scoper.inc.php#L48. It would be great to fix it here at the core rather than those ugly brittle patches

nikic:
@theofidry Done with https://github.com/nikic/PHP-Parser/commit/94ca9a7ab9a9eff7fa15a4a173a8735755ed30f8.

theofidry:
Awesome, thanks :)

PR #444, opened 2017-11-16
Avoid empty first part for global namespace in NameResolver::resolveC…

VolCh:
…lassName

During investigate https://github.com/goaop/parser-reflection/issues/80 and https://github.com/goaop/parser-reflection/pull/83 I found that root of issue is in `Name::$parts[0] === ''` in global namespace which append by NameResolver::resolveClassName during resolve class names using in global namespace like fetch class const.

### DIFF
--- a/lib/PhpParser/NodeVisitor/NameResolver.php
+++ b/lib/PhpParser/NodeVisitor/NameResolver.php
@@ -215,7 +215,13 @@ class NameResolver extends NodeVisitorAbstract
             return FullyQualified::concat($alias, $name->slice(1), $name->getAttributes());
         }
 
-        // if no alias exists prepend current namespace
+        $isGlobalNamespace = $this->namespace !== null && $this->namespace->toString() === '';
+        if ($isGlobalNamespace) {
+            return new FullyQualified($name, $name->getAttributes());
+        }
+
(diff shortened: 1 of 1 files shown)


nikic:
Do you have a reproduce script for this? This looks like a usage mistake to me, as the parser shouldn't ever generate a namespace for which toString() is `''`.

VolCh:
As far I understand it generated by `new Namespace_(new FullyQualified(''), ...)` at https://github.com/goaop/parser-reflection/blob/master/src/NodeVisitor/RootNamespaceNormalizer.php#L47
during parsing file without any namespace, i. e. global namespaces

nikic:
That's the problem then. The global namespace is represented using `new Namespace(null, ...)`, not `new Namespace_(new FullyQualified(''), ...)`. Generally names with empty components are always invalid. I guess it would be good to add an explicit check + exception in this case.

nikic:
I've added an exception on empty names in master: https://github.com/nikic/PHP-Parser/commit/b507fa43dae5b00e60f1ba9d84f48fc323ab46ae Not changing this in 3.x for BC reasons. It looks like goaop/parser-reflection already has an open PR to fix this issue: https://github.com/goaop/parser-reflection/pull/68

PR #458, opened 2017-12-31
Update license year

m1guelpf:
TO DO:

- [x] Wait to the next year (100% done, depends on timezone)
- [ ] Merge!

:octocat:


nikic:
Merged this as https://github.com/nikic/PHP-Parser/commit/67df02c844b92c40220519fe0363b863a3b09c3f tweaked to use a year range instead.

PR #462, opened 2018-01-10
Clean up PHP-Parser

carusogabriel:
Just some CS changes. One change per commit to facilitating review.

Passing [`?w=1`](https://github.com/nikic/PHP-Parser/pull/462/files?w=1) to the Files tab, the review is easier, as GitHub hides `WhiteSpace only changes` :+1:


TomasVotruba:
What about adding coding standard to set this and forget?

carusogabriel:
@TomasVotruba Is something to consider. But we would need to be careful, there are some classes, like `Php5` that have closures with fold methods and sentences. Unfold them would create an amount of code pharaonic. I did the most simple ones. I guess we can join forces to simplify some classes and then apply a Coding Standard :sweat_smile:

TomasVotruba:
Agreed, of course only case where it makes sense :)

@nikic What do you think? It would make contributions and upgrading code easier

lib/PhpParser/BuilderHelpers.php
```diff
@@ -184,11 +185,11 @@ public static function normalizeValue($value) : Expr {
     public static function normalizeDocComment($docComment) : Comment\Doc {
         if ($docComment instanceof Comment\Doc) {
             return $docComment;
-        } else if (is_string($docComment)) {
+        } elseif (is_string($docComment)) {
             return new Comment\Doc($docComment);
-        } else {
-            throw new \LogicException('Doc comment must be a string or an instance of PhpParser\Comment\Doc');
         }
+            throw new \LogicException('Doc comment must be a string or an instance of PhpParser\Comment\Doc');
```
TomasVotruba: I think indention need to be fixed too

nikic:
I've merged most of the changes in this PR two weeks ago, but forgot to comment here on the rest. I'm not interested in doing changes like forcing single quotes or adding trailing commas. I don't see the benefit in these changes, they will only make it hard to rebase patches.

nikic:
Btw, I think the double quote -> single quote conversion pass is really not just unnecessary, but actively harmful. Looking at https://github.com/nikic/PHP-Parser/pull/462/files#diff-9805ffb46197468914fd023f16202c98R38, I find it very hard to see this as an improvement. It takes code that is self-consistent (everything uses double quotes) and converts it into a weird mix of single and double quotes depending on whether that particular string happens to need an escape sequence or not.

carusogabriel:
@nikic Perfect, I do understand your point, just did to follow the best practices out there :smile: 

Is there any other commit that you are interesting in, or may I close it?

PR #468, opened 2018-01-18
Some fixes from PHPStan

carusogabriel:


lib/PhpParser/BuilderHelpers.php
```diff
@@ -16,6 +16,8 @@
  */
 final class BuilderHelpers
 {
+    private $flags;
```
nikic: Why this property?

lib/PhpParser/BuilderHelpers.php
```diff
@@ -132,7 +134,7 @@ public static function normalizeType($type) {
      *
      * @param Node\Expr|bool|null|int|float|string|array $value The value to normalize
      *
-     * @return Expr The normalized value
+     * @return Node|Expr The normalized value
```
nikic: Why the extra return type? I don't see a non-Expr return value.

lib/PhpParser/ParserAbstract.php
```diff
@@ -102,7 +102,7 @@
     protected $startAttributeStack;
     /** @var array[] End attribute stack */
     protected $endAttributeStack;
-    /** @var array End attributes of last *shifted* token */
+    /** @var array|string End attributes of last *shifted* token */
```
nikic: Why this extra type? I think there is some confusion between 3.x branch and master maybe?

lib/PhpParser/BuilderHelpers.php
carusogabriel: Sorry, miss one from version 3.x, gonna rollback :smile:

lib/PhpParser/ParserAbstract.php
carusogabriel: Yep, I'll review the changes between branches, sorry!

lib/PhpParser/PrettyPrinterAbstract.php
```diff
@@ -381,7 +381,7 @@ protected function pPrec(Node $node, int $parentPrecedence, int $parentAssociati
     protected function pImplode(array $nodes, string $glue = '') : string {
         $pNodes = [];
         foreach ($nodes as $node) {
-            if (null === $node) {
+            if (empty($node)) {
```
carusogabriel: This change was made because of:
```
Strict comparison using === between null and PhpParser\Node will always evaluate to false.
```

lib/PhpParser/PrettyPrinterAbstract.php
```diff
@@ -418,7 +418,7 @@ protected function pCommaSeparatedMultiline(array $nodes, bool $trailingComma) :
         $result = '';
         $lastIdx = count($nodes) - 1;
         foreach ($nodes as $idx => $node) {
-            if ($node !== null) {
+            if (!empty($node)) {
```
carusogabriel: This change was made because of:
```
Strict comparison using !== between PhpParser\Node and null will always evaluate to true.
```

lib/PhpParser/PrettyPrinterAbstract.php
TomasVotruba: I don't understand the message. Oh, maybe it's beacuase only `Node` is expected in `$nodes`.
If so, it should be changed above in annotation for `$nodes` to `Node|null`

nikic: Yes, the type hint should be `(Node|null)[]`. If PhpStorm supports it correctly, I remember that there might be an issue.

PR #477, opened 2018-02-22
Replace string by class call

theofidry:
This ensures this piece of code is still working after changing the namespace of the class like done with [PHP-Scoper](https://github.com/humbug/php-scoper)


lib/PhpParser/NodeAbstract.php
```diff
@@ -25,7 +27,7 @@ public function getType() {
         return strtr(
             substr(
                 $className,
-                strpos($className, 'PhpParser\Node') + 15
+                strpos($className, Node::class) + strlen(Node::class) + 1
```
nikic: In that case the strpos call here should be dropped. It was originally added to deal with the same issue. But this solution (using ::class) looks better.

theofidry:
There you go :)

PR #490, opened 2018-03-26
Fix Name\Relative result

carusogabriel:
Sorry if it is a misunderstood trivial change, but shouldn't `Relative` print 

```diff
-namespace\Foo::class
+Foo::class
```

If not, what should I use to create relatives `::class`? :)

Ref: https://github.com/rectorphp/rector/pull/379


m1guelpf:
@carusogabriel I'm simply using Name for this.

nikic:
As @m1guelpf said, if you want an unqualified reference, just use Name. `Name\Relative` is for explicitly namespace-relative references that bypass the alias table.

PR #494, opened 2018-04-17
Update 2_Usage_of_basic_components.markdown

loekiedepo:
1 function call was missing from the docs, which made the tutorial crash.


loekiedepo:
Yeah it is!

nikic:
Okay, makes sense, this was relying on an implicit `__toString()` cast before.

loekiedepo:
Ah, I was wondering where it came from. Thanks!

PR #495, opened 2018-04-17
Fix short array syntax

theofidry:
Ensures that when the long array syntax is enforced, e.g. for PHP 5.3 compatible code, it is properly picked.

The change can be ported to master as well.

### DIFF
--- a/lib/PhpParser/PrettyPrinter/Standard.php
+++ b/lib/PhpParser/PrettyPrinter/Standard.php
@@ -518,8 +518,14 @@ class Standard extends PrettyPrinterAbstract
     }
 
     protected function pExpr_Array(Expr\Array_ $node) {
-        $syntax = $node->getAttribute('kind',
-            $this->options['shortArraySyntax'] ? Expr\Array_::KIND_SHORT : Expr\Array_::KIND_LONG);
+        if (!$this->options['shortArraySyntax']) {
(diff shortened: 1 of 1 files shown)


nikic:
The behavior here was sort-of intentional with `shortArraySyntax` intended as the default if nothing was explicitly specified (in particular what would be used for newly inserted nodes that don't explicitly specify the attribute.) I would find it weird if a global setting had precedence over per-node settings.

nikic:
The behavior is also documented as such: https://github.com/theofidry/PHP-Parser/blob/d3c6f670ede86d619fca2459deb0397d2c0b90e1/lib/PhpParser/PrettyPrinterAbstract.php#L141

theofidry:
>shortArraySyntax intended as the default if nothing was explicitly specified

This intend is not changed.

My fix is, if use specify the traditional array syntax, this one is forced instead of being picked as the default. If you have the following:

```php
$x = [];
```

`$x` is an  `_Array` node with the short array syntax. Because right now the setting is picked as a default value, the printer will print `[]` regardless of your setting. My fix enforce the choice picked my the setting you putted

theofidry:
Closing as I don't think it's the proper fix issue.

Ideally I would like the Printer to be able to print valid PHP code in a given target regardless of what version the original code was written in. That would however require more changes, e.g. being able to drop some typehints and return values and bail out properly when no equivalent is given.

PR #505, opened 2018-05-27
[WIP] Add reference table of the node types

### LINKED ISSUE #501: Where can I get the reference table of the node type?
This maybe a stupid question, but I read all the documents about this project still cannot find a full referrence table.
Currently I only know that 'Stmt' stands for 'statement' and 'Expr' means 'expression'.

carusogabriel:
Closes #501.

Roadmap:
- [ ] List all Nodes
- [ ] List all subNodes
- [ ] Group same types in different files
- [ ] Describe better each Node Type

### DIFF
--- a/doc/component/node_table.markdown
+++ b/doc/component/node_table.markdown
@@ -1,0 +1,55 @@
+## Expr
+
+Class | Node
+----- | -----
+`PhpParser\Node\Expr\BooleanNot` | `!` operator
+`PhpParser\Node\Expr\Empty_` | `empty()` function
+`PhpParser\Node\Expr\Exit_` | `exit` function
+`PhpParser\Node\Expr\Instanceof_` | `instanceof` operator
+`PhpParser\Node\Expr\Isset_` | `isset` operator
+`PhpParser\Node\Expr\List_` | `list()` language constructor. `[]` syntax is also handled by this node
+`PhpParser\Node\Expr\New_` | `new` keyword
+
+## Expr\Cast
+
+Class | Node
+----- | -----
+`PhpParser\Node\Expr\Cast\Array_` | `(array)`
+`PhpParser\Node\Expr\Cast\Bool_` | `(bool)`
+`PhpParser\Node\Expr\Cast\Double` | `(double)`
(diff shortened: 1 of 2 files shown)


nikic:
[Rendered](https://github.com/carusogabriel/PHP-Parser/blob/reference-table-node-type/doc/component/node_table.markdown)

doc/component/node_table.markdown
```diff
@@ -0,0 +1,55 @@
+## Expr
+
+Class | Node
+----- | -----
+`PhpParser\Node\Expr\BooleanNot` | `!` operator
+`PhpParser\Node\Expr\Empty_` | `empty()` function
+`PhpParser\Node\Expr\Exit_` | `exit` function
+`PhpParser\Node\Expr\Instanceof_` | `instanceof` operator
+`PhpParser\Node\Expr\Isset_` | `isset` operator
+`PhpParser\Node\Expr\List_` | `list()` language constructor. `[]` syntax is also handled by this node
```
nikic: I believe `[] = ` syntax uses `Array_`.

nikic:
Would it make sense to also list the subnodes (and their types)? Or would that make things too unwieldy?

doc/component/node_table.markdown
carusogabriel: Thanks, gonna fix it.

nikic:
By subnodes I meant the properties of the node. So `Class_` has `flags`, `name`, `stmts` etc.

> Gonna try to list all of them this afternoon, so we can start to struct the files.
 
I think it would be best to keep everything in one file, otherwise search is hard. E.g. you'd already have to know whether something is an expression or a statement to search for it.

carusogabriel:
> By subnodes I meant the properties of the node. So Class_ has flags, name, stmts etc.

Interesting didn't think on that. Added to our roadmap.

>  E.g. you'd already have to know whether something is an expression or a statement to search for it.

Good point, so let's keep this in one file for now.

I'm sorry for my delay, I'm a little busy with some stuff from the work, but I hope we can document everything this weekend :)

PR #510, opened 2018-06-25
Remove flags from phpunit.xml that have default value and colorize

carusogabriel:
A little improvement for `phpunit.xml.dist`: remove flags that have the default value [as upstream](https://phpunit.de/manual/6.5/en/appendixes.configuration.html) and colorize the output with `colors=true` :)


TomasVotruba:
Love this!

Is there a way to automate this?

PR #511, opened 2018-07-04
Improvements

carusogabriel:
Together with PHPStorm and https://github.com/kalessil/phpinspectionsea, here are some improvements that we can make to PHP-Parser :smile:

### DIFF
--- a/lib/PhpParser/BuilderHelpers.php
+++ b/lib/PhpParser/BuilderHelpers.php
@@ -126,7 +126,7 @@ final class BuilderHelpers
     private static function normalizeNameCommon($name, bool $allowExpr) {
         if ($name instanceof Name) {
             return $name;
-        } elseif (is_string($name)) {
+        } elseif (\is_string($name)) {
             if (!$name) {
                 throw new \LogicException('Name cannot be empty');
             }
@@ -134,7 +134,7 @@ final class BuilderHelpers
             if ($name[0] === '\\') {
                 return new Name\FullyQualified(substr($name, 1));
             } elseif (0 === strpos($name, 'namespace\\')) {
-                return new Name\Relative(substr($name, strlen('namespace\\')));
+                return new Name\Relative(substr($name, \strlen('namespace\\')));
             } else {
                 return new Name($name);
             }
@@ -163,7 +163,7 @@ final class BuilderHelpers
      * @return Name|Identifier|NullableType The normalized type
      */
     public static function normalizeType($type) {
-        if (!is_string($type)) {
+        if (!\is_string($type)) {
             if (!$type instanceof Name && !$type instanceof Identifier
                     && !$type instanceof NullableType) {
                 throw new \LogicException(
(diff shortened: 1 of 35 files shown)


lib/PhpParser/BuilderHelpers.php
```diff
@@ -126,15 +126,15 @@ public static function normalizeNameOrExpr($name) {
     private static function normalizeNameCommon($name, bool $allowExpr) {
         if ($name instanceof Name) {
             return $name;
-        } elseif (is_string($name)) {
+        } elseif (\is_string($name)) {
```
theofidry: I personally find it horrendous to do that: it makes the core really unreadable besides this optimisation can in theory benefit to _all_ functions (although less than if opcache optimised). So I would _strongly_ suggest to import symbols with use statements, e.g. `use function string` or `use const PHP_EOL` rather than this

lib/PhpParser/Node/Name/FullyQualified.php
```diff
@@ -31,15 +31,6 @@ public function isFullyQualified() : bool {
         return true;
     }
 
-    /**
-     * Checks whether the name is explicitly relative to the current namespace. (E.g. namespace\Name)
-     *
-     * @return bool Whether the name is relative
-     */
-    public function isRelative() : bool {
```
theofidry: I'm not against the change but isn't that a BC break?

lib/PhpParser/Node/Name/Relative.php
```diff
@@ -22,15 +22,6 @@ public function isQualified() : bool {
         return false;
     }
 
-    /**
-     * Checks whether the name is fully qualified. (E.g. \Name)
-     *
-     * @return bool Whether the name is fully qualified
-     */
-    public function isFullyQualified() : bool {
```
theofidry: same here

test/PhpParser/Lexer/EmulativeTest.php
```diff
@@ -63,8 +63,7 @@ public function testLexNewFeatures($code, array $expectedTokens) {
         $lexer = $this->getLexer();
         $lexer->startLexing('<?php ' . $code);
 
-        foreach ($expectedTokens as $expectedToken) {
-            list($expectedTokenType, $expectedTokenText) = $expectedToken;
+        foreach ($expectedTokens as list($expectedTokenType, $expectedTokenText)) {
```
theofidry: `[$expectedTokenType, $expectedTokenText]`?

lib/PhpParser/Node/Name/FullyQualified.php
carusogabriel: Nops. As `FullyQualified` extends the `Node\Name`, `isRelative` method is still available. I only removed because the code in parent's one is the same.

test/PhpParser/Lexer/EmulativeTest.php
carusogabriel: We don't use PHP 7.1 here, _yet_ :smile:

theofidry: damn

PR #515, opened 2018-07-19
Add $type argument to BuilderFactory::use

ProgMiner:


nikic:
An alternative would be to add `useFunction()` and `useConst()` methods, which might be more ergonomic. What do you think?

ProgMiner:
Yes, this is better idea.
I think these additional functions can be called `funcUse` and `constUse` like `funcCall` and `constFetch`. What do you think about it?

nikic:
Merged as https://github.com/nikic/PHP-Parser/commit/53c732a96d5e50b5c93a5b7b8a074731b5411d1d with some tests in https://github.com/nikic/PHP-Parser/commit/b855c1ca237764702a9007b5a6646c2847a48a9c. Thanks!

PR #522, opened 2018-07-21
Add array builder

ProgMiner:
I think the setting array node short kind in builder is a useful feature.


nikic:
I'm wondering if this is not better covered by the `shortArraySyntax` option in the pretty printer (https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/PrettyPrinterAbstract.php#L141)? If you want to use short array syntax, you likely want to use them everywhere, rather than chose separately each time.

ProgMiner:
Oh, thanks. I don't know about this

PR #527, opened 2018-08-07
Failing test case for PrettyPrinter::printFormatPreserving() - lost off line breaks

weaverryan:
Hi there!

I've *finally* found one small situation where the `printFormatPreserving()` is not *quite* preserving all of the original code/whitespace :). This PR is just the failing test case, as I'm honestly not sure where to look to fix it. Here is the output from the failing test:

```diff
--- Expected
+++ Actual
@@ @@
 '<?php
 interface Foo {}
 
-class Bar implements Foo {
+class Bar implements Foo
+{
     private $fooProp;
-
     public function getFooProp()
     {
         return $this->fooProp;
     }
 }'
```

The movement of the `{` from the class line to the next line is not a huge deal - after all, we *are* modifying this line. The bigger problem is the lost line break between the `$fooProp` property and `getFooProp()`. Actually, *all* line breaks are lost inside the class body. It looks like the modified interface causes `printFormatPreserving()` to think that it needs to re-pretty-print all of the class body.

Do you have any hints on where to look to fix this? Easy fix? Or is this more involved?

Thanks!

### DIFF
--- a/test/code/formatPreservation/addInterface.test
+++ b/test/code/formatPreservation/addInterface.test
@@ -1,0 +1,27 @@
+Add interface
+-----
+<?php
+interface Foo {}
+
+class Bar {
+    private $fooProp;
+
+    public function getFooProp()
+    {
+        return $this->fooProp;
+    }
+}
+-----
+$stmts[1]->implements[] = new Node\Name('Foo');
+-----
+<?php
+interface Foo {}
+
+class Bar implements Foo {
+    private $fooProp;
+
+    public function getFooProp()
+    {
+        return $this->fooProp;
+    }
+}


nikic:
That would be due to https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/PrettyPrinterAbstract.php#L874. Unfortunately not an easy fix. Though just special-casing the "implements" case might be possible.

weaverryan:
Thanks for the info! Unfortunately, it's still over my head :/. I've worked around it on our end (no big deal). If you don't think it's worth fixing, then please feel free to close.

Cheers!

nikic:
It's definitely worth fixing, but may take a while :)

nikic:
That "a while" turned into a very long while, but I finally added support for this: https://github.com/nikic/PHP-Parser/commit/aa72c5d67478f56d0365195a8bc5e14df8b6540d

weaverryan:
Wow! Thank you!

TomasVotruba:
@weaverryan I'm curious, where do you use PHP parser? Any opensource to share?

weaverryan:
MakerBundle - we use it there to update existing source code without changing their classes more than we need to.

PR #528, opened 2018-08-13
Remove NodeTraverser dependency on parent::__construct()

TomasVotruba:
At the moment, when `NodeTraverser` is extended and constructor used for dependency injection, there is hidden and unneded dependency on `parent__construct()` call:

```php
use PhpParser\NodeTraverser;

final class NodeTypeResolverNodeTraverser extends NodeTraverser
{
    public function __construct(MetadataNodeVisitor $metadataNodeVisitor)
    {
        // this would crash, since property is `null`
        foreach ($this->visitors ...) {
        }
        // ...
    }
}
```

Using `[]` as default value would prevent similar bugs.

### DIFF
--- a/lib/PhpParser/NodeTraverser.php
+++ b/lib/PhpParser/NodeTraverser.php
@@ -31,16 +31,13 @@ class NodeTraverser implements NodeTraverserInterface
     const REMOVE_NODE = 3;
 
     /** @var NodeVisitor[] Visitors */
-    protected $visitors;
+    protected $visitors = [];
 
     /** @var bool Whether traversal should be stopped */
     protected $stopTraversal;
 
-    /**
-     * Constructs a node traverser.
-     */
     public function __construct() {
-        $this->visitors = [];
+        // for BC
     }
 
     /**


GrahamCampbell:
Isn't this a breaking change? Calls to the parent construction will now throw an error after this change?

image

TomasVotruba:
What do you suggest then?

GrahamCampbell:
Not merging this PR, because it's a breaking change.

TomasVotruba:
:D I mean to change to not make it BC break.

theofidry:
@TomasVotruba having an empty constructor would do:

https://3v4l.org/1hkph

Maybe a `@trigger_error()` deprecation message could be added to warn people to no longer call `parent::__construct()`

nikic:
I don't really see the point of this change.

TomasVotruba:
I managed to foreach `$visitors` with null many times already. This change would prevent that.

There is no real need to remove constructor, but rather to add default value for public iterable variable of class that can be extened.

@theofidry Thanks for good point!

nikic:
Okay, I'm fine with the change as long as the empty ctor remains, for the mentioned BC reasons.

TomasVotruba:
Thank you :+1:

ondrejmirtes:
@TomasVotruba You should always call parent constructor. Even PHPStan points that out 😊

Issue #534, opened 2018-09-25
Do not visit children and the current node for subsequent visitors


borNfreee:
This is an attempt to reopen https://github.com/nikic/PHP-Parser/issues/164 but with a real world case rather than "just for performance purposes".

We are writing a Mutation Testing framework, where each Node can be mutated to one or more other nodes, or removed completely. Small example for reference:

```diff
- public function execute()
+ protected function execute()
```

Then, when developer uses `@codeCoverageIgnore` on the method level, we should not mutate anything and just skip it from the mutation process:

```php
class Executor
{
    /**
     * @codeCoverageIgnore
     */
    public function execute()
    {
        // ...
    }
}
```

To achieve this, we can write a simple `NodeVisitor`:

```php
final class CodeCoverageMethodIgnoreVisitor extends NodeVisitorAbstract
{
    public function enterNode(Node $node)
    {
        if (!$node instanceof Node\Stmt\ClassMethod) {
            return null;
        }

        $method = new \ReflectionMethod(...);

        $docComment = $method->getDocComment();

        if ($docComment === false) {
            return null;
        }

        if (strpos($docComment, '@codeCoverageIgnore') !== false) {
           return NodeTraverser::DONT_TRAVERSE_CHILDREN;
        }

        return null;
    }
}
```

This works great, except that the node being visited is still mutated. That is expected, because `Traverser` does not visit *children*, but not the node itself.

And since we have such sequense of visitors:

```php
$traverser->addVisitor(new CodeCoverageMethodIgnoreVisitor());
$traverser->addVisitor(new MutationCollector());
```

`MutationCollector` visits the same `ClassMethod` node and collects mutation.

So, the question is: is it possible to add one more constant that will work exaclty like `DONT_TRAVERSE_CHILDREN`, but in addition won't mutate *current* node (and yes, this will require additinal logic for `leaveNode()` visitors).

Thank you.

nikic:
I'm fine with adding the extra flag, though I'd appreciate a PR for this.

PR #536, opened 2018-10-02
Add new constant to be returned from enterNode() to not travers current and child nodes

borNfreee:
Detailed explanation in https://github.com/nikic/PHP-Parser/issues/534

I'm not sure about naming, so any advice is highly appreciated.

Fixes https://github.com/nikic/PHP-Parser/issues/534

### DIFF
--- a/lib/PhpParser/NodeTraverser.php
+++ b/lib/PhpParser/NodeTraverser.php
@@ -30,6 +30,15 @@ class NodeTraverser implements NodeTraverserInterface
      */
     const REMOVE_NODE = 3;
 
+    /**
+     * If NodeVisitor::enterNode() returns DONT_TRAVERSE_CURRENT_AND_CHILDREN, child nodes
+     * of the current node will not be traversed for any visitors.
+     *
+     * For subsequent visitors enterNode() will not be called as well.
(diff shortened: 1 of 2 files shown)


lib/PhpParser/NodeTraverser.php
```diff
@@ -134,8 +149,13 @@ protected function traverseNode(Node $node) : Node {
                     }
                 }
 
-                foreach ($this->visitors as $visitor) {
+                foreach ($this->visitors as $visitorIndex => $visitor) {
                     $return = $visitor->leaveNode($subNode);
+
+                    if ($breakVisitorIndex === $visitorIndex) {
```
nikic: This should probably come after the checks below, otherwise the visitor that returned DONT_TRAVERSE_CURRENT_AND_CHILDREN  in enterNode will not be able to perform a replacement in leaveNode.

borNfreee: I've updated the code and test to make sure `leaveNode()` can replace the node when `DONT_TRAVERSE_CURRENT_AND_CHILDREN` is used

nikic:
Thanks, looks great!

borNfreee:
Thank you!

PR #537, opened 2018-10-03
Add prioritised traversal of nodes

BackEndTea:


lib/PhpParser/NodeTraverser.php
```diff
@@ -49,6 +49,22 @@ public function addVisitor(NodeVisitor $visitor) {
         $this->visitors[] = $visitor;
     }
 
+    /**
+     *
```
borNfreee: extra line?

nikic:
As this is both easy to implement outside of PHP-Parser and not commonly needed functionality, I'm skeptical about including this. If I were designing this class today I'd drop addVisitor/removeVisitor altogether and only accept an array on construction -- the job of this class is to traverse an AST, not to reimplement an array or priority queue...

TomasVotruba:
I also use priorities and now do it like this:

https://github.com/rectorphp/rector/blob/master/packages/NodeTypeResolver/src/NodeScopeAndMetadataDecorator.php#L83-L90

Accepting on construction would be more clear.
Priorities on the other hand would allow making use of DI container. It's common pattern to define priority inside the service and let DI handle that.

lib/PhpParser/NodeTraverser.php
```diff
@@ -49,6 +49,21 @@ public function addVisitor(NodeVisitor $visitor) {
...
+     * @param NodeVisitor $visitor
+     * @param int $priority
+     */
+    public function addPrioritizedVisitor(NodeVisitor $visitor, int $priority) {
+        if(\array_key_exists($priority, $this->visitors)) {
+            throw new \LogicException(sprintf('Priority %s already exists.', $priority));
+        }
+
+        $this->visitors[$priority] = $visitor;
+        krsort($this->visitors);
```
TomasVotruba: Very elegant solution btw. I always have extra "sortByPriority()" method and have to call it every where. This is much better!

PR #546, opened 2018-10-28
add php 7.3 to travis

andreybolonin:


nikic:
We need to wait until Travis updates to PHP 7.3 RC4, which has a bug fix we need (https://bugs.php.net/bug.php?id=76991).

nikic:
Restarted builds, Travis is still on an old PHP 7.3 version :(

GrahamCampbell:
Travis is on RC5 now.

PR #548, opened 2018-11-13
Exclude files and folders from release archive

ankurk91:
Hi,

Let's not publish all those file.
`.gitattributes` can be used to exclude files from final archive that is downloaded by composer.

Read more
- https://madewithlove.be/gitattributes/
- https://alexbilbie.com/2012/11/exclude-objects-with-gitattributes/
- http://www.pixelite.co.nz/article/using-git-attributes-exclude-files-your-release/

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+/doc export-ignore
+/test export-ignore
+/test_old export-ignore
+/.gitignore export-ignore
+/.travis.yml export-ignore
+/phpunit.xml.dist export-ignore


ankurk91:
@nikic 
Am i missing something here?

nikic:
See #178. Just to double check, @remicollet, you still feel the same way about not excluding any files via .gitattributes?

remicollet:
> you still feel the same way about not excluding any files via .gitattributes?

Yes

(sadly PHP is the only language without a proper installer, able to manage file role, such as the old pear/pecl, npm, cpan, pip, ...)

PR #549, opened 2018-11-20
[cs] remove unused @var name

TomasVotruba:


nikic:
Is this the only occurrence of this issue?

nikic:
Thanks!

PR #553, opened 2018-11-30
Add lexer error when encountering invalid strings

### LINKED ISSUE #551: Undefined offset when parsing malformed string in array
```php
$a = ["a "thing"];
```

Output is:
```
Undefined offset: 6
.../vendor/nikic/php-parser/lib/PhpParser/Parser/Php7.php:2361
```

muglug:
Fixes #551

Let me know it would unduly affect performance or introduce regressions.

Thanks for all your work on PHP Parser!


nikic:
I haven't looked into this issue yet, but usually "undefined offset" errors are a result of a bug in the generated parser or the parsing algorithm. It may be that adding these extra productions fixes this particular case, but the issue may appear through some other way. I'll try to look at some parser traces tomorrow.

muglug:
A different alternative fix means that error recovery is diminished for
```php
[$foo $bar]; // only $foo is in parsed results
```

muglug:
I think the culprit is

```
comma_or_error:
      ','
    | error
;
```

Removing `| error` prevents the problem (but also breaks more recovery tests than my fix)

nikic:
You're right that `comma_or_error` is at fault here. I pushed a different fix at f42bbc2403ebc1178d9ae4f1029c0e8ddc6aa451 which simply removes the code generating the notice.

If no explicit action for a production is given, it defaults to performing `$$=$1`, which does not make a lot of sense for `error`, which will generally not have any meaningful value. This change just suppresses this default.

muglug:
Awesome, thanks!

PR #562, opened 2018-12-25
[WIP] Fix trialing comma for array

zhujinxuan:
Work in progress


zhujinxuan:
Hi, @nikic . it is almost done.  But there is still a problem, how can I get the line number of all commas in checkArray_ ?  Currently, I have no idea how to access the line number.

lib/PhpParser/ParserAbstract.php
```diff
@@ -824,6 +824,12 @@ protected function parseDocString(
         }
     }
 
+    protected function checkArray_(Node\Expr\Array_ $a) {
```
nikic: The trailing `_` here is not necessary. It's only used in `Array_` to avoid clashing with the reserved keyword.

lib/PhpParser/ParserAbstract.php
```diff
@@ -824,6 +824,12 @@ protected function parseDocString(
...
+        if (in_array(null, $a->items, true)) {
```
nikic: This will also have to check the array recursively to catch cases like `$x = [1, [2,,,,]];`.

zhujinxuan:
Hi, @nikic .  I have finished on the comment and emit error with start line.  It is ready for review now.

zhujinxuan:
I think we shall have separate array_item_left and array_short_list_left, the following code is illegal in php but legal in current parser:

```
array($a)] = [[1]];
```

zhujinxuan:
Cannot be fixed by adding `expr_left`.  kmycc consider the definition of `expr_left` and `expr` conflicting because they both have variable and '['.

zhujinxuan:
Cannot find a way to fix it.  Close this PR

PR #565, opened 2019-01-03
Add float/real/double kind to Expr\Cast\Double_

### LINKED ISSUE #564: Ability to differntiate (real), (float) and (double)
I know these are technically synonymous, but it would be handy to detect them.
Why? Becaus `(real)` case is [about to be deprecated in PHP 7.4](https://wiki.php.net/rfc/deprecations_php_7_4#the_real_type).

Same as `Array_` has:

```php
$items = [];
$items = array();
```

```php
class Array_ extends Expr
{
    const KIND_LONG = 1;  // array() syntax
    const KIND_SHORT = 2; // [] syntax
```

E.g.

```php
class Double extends Cast
{
    const KIND_DOUBLE = 1;
    const KIND_FLOAT = 3;
    const KIND_REAL = 3;
```

What do you think?

TomasVotruba:
Closes #564 

This is my first real contribution to the parser, I used [the fork README reference](https://github.com/nikic/PHP-Parser/tree/master/grammar#building-the-parser) and learned present syntax.

Let me know if I missed something.


grammar/php5.y
```diff
@@ -628,7 +628,17 @@ expr:
     | T_REQUIRE expr                                        { $$ = Expr\Include_[$2, Expr\Include_::TYPE_REQUIRE]; }
     | T_REQUIRE_ONCE expr                                   { $$ = Expr\Include_[$2, Expr\Include_::TYPE_REQUIRE_ONCE]; }
     | T_INT_CAST expr                                       { $$ = Expr\Cast\Int_    [$2]; }
-    | T_DOUBLE_CAST expr                                    { $$ = Expr\Cast\Double  [$2]; }
+    | T_DOUBLE_CAST expr
+          { $attrs = attributes();
+            $cast = strtolower($1);
+            if ($cast === '(float)') {
```
nikic: Things like `(  float  )` should also be handled.

TomasVotruba: Sure. Any tips how? Regular expression?

nikic: Maybe only check if it contains float/real/double, rather than exact equality?

TomasVotruba: Done :+1:

grammar/php5.y
```diff
@@ -628,7 +628,17 @@ expr:
...
+    | T_DOUBLE_CAST expr
+          { $attrs = attributes();
+            $cast = strtolower($1);
+            if (strpos($cast, 'float') !== false) {
+                $attrs['kind'] = Expr\Cast\Double::KIND_FLOAT;
+            } elseif (strpos($cast, 'real') !== false) {
+                $attrs['kind'] = Expr\Cast\Double::KIND_REAL;
+            } else {
+                $attrs['kind'] = Expr\Cast\Double::KIND_DOUBLE;
+            }
```
nikic: Can you please extract this into a method on `ParserAbstract` to avoid the code duplication? `$this->getCastKind($1)` or similar.

TomasVotruba: I'll look into it

TomasVotruba: Cool architecture, easy to understand :+1:

TomasVotruba:
I'll squash commits to 1

PR #567, opened 2019-01-05
[PHP 7.4] Add typed properties

TomasVotruba:
[PHP 7.4 typed properties RFC is accepted](https://wiki.php.net/rfc/typed_properties_v2). Now I have to [use attributes](https://github.com/rectorphp/rector/pull/643/files#diff-d340c7df540adc4b55adb245a50fcc00R50) to add PHP 7.4 property types. Why not add it here? 

I inspired by [return type declaration](https://github.com/nikic/PHP-Parser/commit/3ad0d4b310a5d61e53a6bb359ff19a8d76fb556a) commit, but I got stuck at parsing. I gave it 3 hours of work but I don't know where is the catch. 

I'll be grateful for feedback. Thank you


grammar/php7.y
```diff
@@ -535,8 +540,8 @@ class_statement_list:
 ;
 
 class_statement:
-      variable_modifiers property_declaration_list ';'
-          { $$ = Stmt\Property[$1, $2]; $this->checkProperty($$, #1); }
+      variable_modifiers optional_property_type property_declaration_list ';'
```
TomasVotruba: I think `variable_modifiers` also eats `optional_property_type`.

I tried `method_modifiers optional_property_type T_VARIABLE property_declaration_list` but without success.

test/code/parser/stmt/class/propertyTypes.test
```diff
@@ -0,0 +1,58 @@
...
+                            name: c
+                        )
+                        default: null
+                    )
+                )
+                type: null
+            )
+        )
+    )
+)
```
TomasVotruba: This test should include types, now `null`

test/code/parser/stmt/class/propertyTypes.test
```diff
@@ -0,0 +1,58 @@
...
+-----
+<?php
+
+class A {
+    public string $a;
+    protected D $b;
+    private ?float $c;
+}
+-----
+array(
```
nikic: Add `!!php7` before this line.

grammar/php7.y
```diff
@@ -535,8 +540,8 @@ class_statement_list:
...
+          { $$ = Stmt\Property[$1, $3, [], $2]; $this->checkProperty($$, #1); }
```
nikic: The `[]` should be `attributes()`.

nikic: More precisely, the whole thing should be `new Stmt\Property($1, $3, attributes(), $2)`.

lib/PhpParser/Builder/Property.php
```diff
@@ -95,6 +103,20 @@ public function setDocComment($docComment) {
         return $this;
     }
 
+    /**
+     * Sets the property type for PHP 7.4+.
+     *
+     * @param string|Name|NullableType $type One of array, callable, string, int, float,
+     *                                       bool, iterable, or a class/interface name.
```
nikic: This list is outdated, let's be less specific here. Also `Identifier` is allowed here.

lib/PhpParser/Node/Stmt/Property.php
```diff
@@ -3,29 +3,36 @@
...
     /**
      * Constructs a class property list node.
      *
-     * @param int                $flags      Modifiers
-     * @param PropertyProperty[] $props      Properties
-     * @param array              $attributes Additional attributes
+     * @param int                           $flags      Modifiers
+     * @param PropertyProperty[]            $props      Properties
+     * @param array                         $attributes Additional attributes
+     * @param null|string|Name|NullableType $type       Typehint
```
nikic: Identifier is also allowed here. Instead of "Typehint" better just "Type" or "Type declaration".

lib/PhpParser/Node/Stmt/Property.php
```diff
@@ -3,29 +3,36 @@
...
+    public function __construct(int $flags, array $props, array $attributes = [], $type = null) {
         parent::__construct($attributes);
         $this->flags = $flags;
         $this->props = $props;
+        $this->type = \is_string($type) ? new Identifier($type) : $type;
     }
 
     public function getSubNodeNames() : array {
-        return ['flags', 'props'];
+        return ['flags', 'props', 'type'];
```
nikic: `type` should come before `props` here. These are in the same order as in the source code.

lib/PhpParser/NodeVisitor/NameResolver.php
```diff
@@ -94,6 +94,10 @@ public function enterNode(Node $node) {
                   || $node instanceof Expr\Closure
         ) {
             $this->resolveSignature($node);
+        } elseif ($node instanceof Stmt\Property) {
+            if ($node->type instanceof Name) {
```
nikic: This should use `resolveType()`. Tests should checks that something like `public ?Foo $bar` is correctly resolved.

grammar/php7.y
TomasVotruba: Oh, so attributes have to be always set like this if not default value?

nikic:
Missing a pretty-printer test (or add to an existing one).

Missing support for adding/removing property types in formatting preserving pretty printer, but I think this is going to be quite hard to do, at least for the insert case.

lib/PhpParser/Node/Stmt/Property.php
TomasVotruba: I tried to respect argument order, so it's consistent inside the code

nikic: The order here is important for pretty printing (see failing test), because it assumes that subnodes never occur out of order. Matching the order in the ctor isn't necessary (and also not done in all other cases -- often the required parts are not necessarily first in code order).

TomasVotruba: I see :+1:

grammar/php7.y
nikic: The `Stmt\Property[$x]` syntax is short for `new Stmt\Property($x, attributes())`. So if you manually pass attributes, you should write the `new` out explicitly. Otherwise it will be converted into `new Stmt\Property($1, $3, attributes(), 2, attributes())`, which PHP allows but is also redundant.

TomasVotruba:
Comments resolved

> Missing a pretty-printer test (or add to an existing one).

I'm on it

grammar/php7.y
TomasVotruba: Thank you

TomasVotruba:
> Missing support for adding/removing property types in formatting preserving pretty printer, but I think this is going to be quite hard to do, at least for the insert case.

Could you link me to test file to get inspiration for this?

TomasVotruba:
...I think I found the test case.

TomasVotruba:
Could you point me to the place in Printer where this should be handled?

test/code/formatPreservation/addingPropertyType.test
```diff
@@ -0,0 +1,35 @@
...
+}
+-----
+$stmts[0]->stmts[0]->type = new Node\Identifier('int');
+-----
+<?php
+
+class A
+{
+    public int
+        $b;
```
TomasVotruba: Here the output actually is:

```php
public int $b;
```

Incorret

test/code/formatPreservation/removingPropertyType.test
```diff
@@ -0,0 +1,17 @@
+Removing property type
+-----
+<?php
+
+class B
+{
+    public ?float $b;
```
nikic: This should have non-standard formatting somewhere, otherwise you can't see whether formatting is preserved or not.

```
public ?float $b
    = 1.0;
```

maybe.

nikic: `Stmt_Property->type` needs to be added to https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/PrettyPrinterAbstract.php#L1184 using `$stripRight` to make this work correctly.

test/code/formatPreservation/addingPropertyType.test
nikic: `Stmt_Property->type` needs to be added to https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/PrettyPrinterAbstract.php#L1224, but I don't think it's going to work out of the box. You can try adding `[null, null, ' ']` and see what happens.

test/code/formatPreservation/removingPropertyType.test
TomasVotruba: Almost there, just type and modifiers are switched:

```php
int public $b
    = 1.0
```

nikic: Okay, that's about what I expected. Instead of using `[null, null, ' ']` we need `[\T_VARIABLE, null, ' ']` plus a way to avoid the `+ 1` here: https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/PrettyPrinterAbstract.php#L622 That is, we want to insert before the token rather than after it.

For example, we could change the code to:

```
                list($findToken, $beforeToken, $extraLeft, $extraRight) = $this->insertionMap[$key];
                if (null !== $findToken) {
                    $subStartPos = $this->origTokens->findRight($pos, $findToken)
                                 + (int) !$beforeToken;
                } else {
                    $subStartPos = $pos;
                }
```

And change the format of the data array to `[\T_VARIABLE, true, null, ' ']`. (and `[x, false, x, x]` for all others). It's ugly, but I don't see a much better way.

test/PhpParser/NodeVisitor/NameResolverTest.php
```diff
@@ -96,6 +96,13 @@ public function testResolveNames() {
     E;
     K;
 }
+namespace BarBaz {
+    class ClassWithTypeProperties
+    {
+        public float $php = 7.4;
+        public Foo $person;
```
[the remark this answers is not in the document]
TomasVotruba: Sure :+1:

grammar/php7.y
```diff
@@ -480,6 +480,11 @@ optional_return_type:
     | ':' error                                             { $$ = null; }
 ;
 
+optional_property_type:
+      /* empty */                                           { $$ = null; }
+    | type_expr                                             { $$ = $1; }
+;
```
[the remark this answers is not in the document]
TomasVotruba: Good point. Done :+1:

test/PhpParser/NodeVisitor/NameResolverTest.php
```diff
@@ -95,6 +95,13 @@ public function testResolveNames() {
     C;
     E;
     K;
+
+    class ClassWithTypeProperties
+    {
+        public float $php = 7.4;
+        public ?Foo $person;
+        protected static ?bool $probability;
+    }
```
TomasVotruba: I accidentaly removed this test case during rebase on `master`.

Is this complete enough?

nikic:
Merged as https://github.com/nikic/PHP-Parser/commit/af8c729603cd972b277697d5cb3f36ef3f10ca76. Thanks!

PR #570, opened 2019-01-09
Enhancement: Reference phpunit.xsd as installed with composer

localheinz:
This PR

* [x] references `phpunit.xsd` as installed with `composer`

💁‍♂️ Helps with auto-completion.


localheinz:
Thank you, @nikic!

PR #571, opened 2019-01-09
Enhancement: Normalize composer.json

localheinz:
This PR

* [x] normalizes `composer.json`

💁‍♂️ I ran

```
$ composer global require localheinz/composer-normalize
```

followed by

```
$ composer normalize
```

For reference, see https://github.com/localheinz/composer-normalize.


localheinz:
Thank you, @nikic!

PR #573, opened 2019-01-12
updates via "rectorphp/rector"

voku:
- "global" -> remove unused "use" statements
- "phpunit" -> fix "@covers" comments
- "phpunit" -> use Iterator for "@dataProvider" (if possible)
- "phpunit" -> replace "->will($this->returnValue()" with "->willReturn()"
- "phpunit" -> replace "->will($this->throwException()" with "->willThrowException()"
- "UseTest.php" -> add missing namespace
- "composer.json" -> use "autoload-dev"

-> most of the changes are done automatically by "https://github.com/rectorphp/rector"

---
This change is [Reviewable](https://reviewable.io/reviews/nikic/php-parser/573)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,11 @@
             "PhpParser\\": "lib/PhpParser"
         }
     },
+    "autoload-dev": {
+        "psr-4": {
+            "PhpParser\\": "test/PhpParser/"
+        }
+    },
     "bin": [
         "bin/php-parse"
     ]
--- a/test/PhpParser/Builder/PropertyTest.php
+++ b/test/PhpParser/Builder/PropertyTest.php
@@ -91,56 +91,57 @@ class PropertyTest extends \PHPUnit\Framework\TestCase
         $this->assertEquals($expectedValueNode, $node->props[0]->default);
     }
 
-    public function provideTestDefaultValues(): \Iterator
-    {
-        yield [
-            null,
-            new Expr\ConstFetch(new Name('null'))
-        ];
-        yield [
-            true,
-            new Expr\ConstFetch(new Name('true'))
-        ];
-        yield [
-            false,
-            new Expr\ConstFetch(new Name('false'))
-        ];
-        yield [
-            31415,
-            new Scalar\LNumber(31415)
-        ];
-        yield [
-            3.1415,
-            new Scalar\DNumber(3.1415)
-        ];
-        yield [
-            'Hallo World',
-            new Scalar\String_('Hallo World')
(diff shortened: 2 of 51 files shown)


test/PhpParser/Builder/PropertyTest.php
```diff
@@ -92,57 +91,56 @@ public function testDefaultValues($value, $expectedValueNode) {
...
-                    ),
-                ])
-            ],
-            [
-                new Scalar\MagicConst\Dir,
-                new Scalar\MagicConst\Dir
-            ]
+    public function provideTestDefaultValues(): \Iterator
+    {
+        yield [
```
nikic: The way PHPUnit works, using generators for data providers unfortunately doesn't make much sense. They will end up materialized as arrays anyway.

test/PhpParser/CodeTestAbstract.php
```diff
@@ -2,11 +2,9 @@
 
 namespace PhpParser;
 
-use PHPUnit\Framework\TestCase;
-
 require_once __DIR__ . '/CodeTestParser.php';
```
nikic: With the autoload-dev change, possibly these requires can be dropped?

nikic:
Not a fan of the `yield` changes, but otherwise this looks good.

test/PhpParser/CodeTestAbstract.php
voku: correct :) these "require_once" usages in the tests have now been removed

nikic:
Thanks!

test/PhpParser/ParserTest.php
```diff
@@ -6,9 +6,8 @@
 use PhpParser\Node\Scalar;
 use PhpParser\Node\Scalar\String_;
 use PhpParser\Node\Stmt;
-use PHPUnit\Framework\TestCase;
 
-abstract class ParserTest extends TestCase
+abstract class ParserTest extends \PHPUnit\Framework\TestCase
```
TomasVotruba: These extends should be impoted with coding standrd run, either `--with-style` or standalone.
Just to be consistent with rest of the imports.

composer.json
```diff
@@ -29,6 +29,11 @@
             "PhpParser\\": "lib/PhpParser"
         }
     },
+    "autoload-dev": {
+        "psr-4": {
+            "PhpParser\\": "test/PhpParser/"
+        }
+    },
```
TomasVotruba: :+1:

PR #575, opened 2019-01-22
[PHP 7.4] Add coalesce equal `??=`

TomasVotruba:
Adds `??=` support https://github.com/php/php-src/pull/3747

Tests are failing.

It looks like it's not possible to add it. Probably needs to wait for `T_COALESCE_EQUAL` token, or not?


nikic:
Yes, it should use `T_COALESCE_EQUAL`. If we run on a PHP version that does not have it, then we need to emulate it in Lexer\Emulative, by combining a T_COALESCE and a `'='` token into a `T_COALESCE_ASSIGN` token.

TomasVotruba:
Thanks for navigation. I dig through history and blame and tried to inspire at: https://github.com/nikic/PHP-Parser/commit/251e6892830568d536fc8eaae6eeb86ab940a0ef

Emulative parsers seems to work, but grammer still looks broken. Any clue?

lib/PhpParser/PrettyPrinterAbstract.php
```diff
@@ -82,6 +82,7 @@
         AssignOp\ShiftLeft::class      => [160,  1],
         AssignOp\ShiftRight::class     => [160,  1],
         AssignOp\Pow::class            => [160,  1],
+        AssignOp\Coalesce::class       => [160,  1],
```
TomasVotruba: Not sure what these numbers stand for, just blind shot

TomasVotruba:
Now the parser is rebuild without error, that good.

Yet the error still remains:

```bash
1) PhpParser\PrettyPrinterTest::testPrettyPrint with data set "var/www/PHP-Parser/test/code/prettyPrinter/expr/operators_php7.test#0" ('Basic operators (/var/www/PHP....test)', '<?php\n\n$a ??= $c;', '$a ??= $c;', 'php7')
RangeException: The lexer returned an invalid token (id=281, value=??=)
```

Test case:

```
Basic operators
-----
<?php

$a ??= $c;
-----
!!php7
$a ??= $c;
```

TomasVotruba:
Any tips what tests should I add?

nikic:
Okay, I believe this issue here is that the token is not used in the PHP 5 parser and thus gets dropped from the translation table. It would be fine if you just add support for `??=` to the PHP 5 parser as well.

One thing that's missing here is a parser test for `??=`. If `??=` is supported in the PHP 5 parser as well, you can just add it into test/code/parser/expr/assign.test.

TomasVotruba:
>  It would be fine if you just add support for ??= to the PHP 5 parser as well.

Works like a charm!

> One thing that's missing here is a parser test for ??=. If ??= is supported in the PHP 5 parser as well, you can just add it into test/code/parser/expr/assign.test.

:+1: I'm on it

lib/PhpParser/Lexer/Emulative.php
```diff
@@ -171,7 +240,8 @@ private function fixupTokens() {
      *
      * @param Error[] $errors
      */
-    private function fixupErrors(array $errors) {
+    private function fixupErrors(array $errors)
```
nikic: Unnecessary formatting change. This project does not follow the "bad parts" of PSR-2 ;)

lib/PhpParser/PrettyPrinterAbstract.php
```diff
@@ -82,6 +82,7 @@ abstract class PrettyPrinterAbstract
...
```
nikic: It should also be added here: https://github.com/nikic/PHP-Parser/blob/716e827d7f10c8cee7760015ad76b954205e5aaa/lib/PhpParser/PrettyPrinterAbstract.php#L1156

nikic:
Merged as https://github.com/nikic/PHP-Parser/commit/9de96821f72851b4972abf417e24e33936106e2a. Thanks :)

TomasVotruba:
:+1:  Thank you

TomasVotruba:
Any ETA for release of this? I'd love to test in Rector and real code to demonstrace PHP 5.3 → 7.4 upgrade

nikic:
Done: https://github.com/nikic/PHP-Parser/releases/tag/v4.2.1

TomasVotruba:
Thank you :heart:

PR #577, opened 2019-01-23
Add attribute of raw string to Scalar String_

### LINKED ISSUE #576: Get unparsed string from String_ for escaped sequences
I'm currently using PHPStan, which uses PHP7 parser provided by `PHP-Parser`.

My objective here is to get unparsed string argument from `String_` class such that I'll be able to detect if there's any escaped sequence in the string.

e.g. I wish to get argument of value `'This is a string with escape sequence \x41'` from `String_` class after parsing the code below.
```php
<?php
Foo::bar("This is a string with escape sequence \x41");
```
However, what I get instead is 
```
'This is a string with escape sequence A'
```
This is because the raw string was parsed.
I wonder if it is possible to get an unparsed string from `String_` class?

Billz95:
Fixes #576

### DIFF
--- a/lib/PhpParser/Node/Scalar/String_.php
+++ b/lib/PhpParser/Node/Scalar/String_.php
@@ -16,6 +16,9 @@ class String_ extends Scalar
     /** @var string String value */
     public $value;
 
+    /** @var string Unparsed String value */
+    public $rawValue;
+
     protected static $replacements = [
         '\\' => '\\',
         '$'  =>  '$',
--- a/lib/PhpParser/Parser/Php7.php
+++ b/lib/PhpParser/Parser/Php7.php
@@ -2197,7 +2197,8 @@ class Php7 extends \PhpParser\ParserAbstract
             },
             437 => function ($stackPos) {
                  $attrs = $this->startAttributeStack[$stackPos-(1-1)] + $this->endAttributes; $attrs['kind'] = ($this->semStack[$stackPos-(1-1)][0] === "'" || ($this->semStack[$stackPos-(1-1)][1] === "'" && ($this->semStack[$stackPos-(1-1)][0] === 'b' || $this->semStack[$stackPos-(1-1)][0] === 'B')) ? Scalar\String_::KIND_SINGLE_QUOTED : Scalar\String_::KIND_DOUBLE_QUOTED);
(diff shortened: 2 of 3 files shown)


lib/PhpParser/Node/Scalar/String_.php
```diff
@@ -16,6 +16,9 @@ class String_ extends Scalar
     /** @var string String value */
     public $value;
 
+    /** @var string Unparsed String value */
+    public $rawString;
```
TomasVotruba: `$rawValue` would be better, to be consistent with `$value` above

lib/PhpParser/Parser/Php7.php
```diff
@@ -2173,6 +2173,7 @@ protected function initReduceCallbacks() {
             433 => function ($stackPos) {
                  $attrs = $this->startAttributeStack[$stackPos-(1-1)] + $this->endAttributes; $attrs['kind'] = ($this->semStack[$stackPos-(1-1)][0] === "'" || ($this->semStack[$stackPos-(1-1)][1] === "'" && ($this->semStack[$stackPos-(1-1)][0] === 'b' || $this->semStack[$stackPos-(1-1)][0] === 'B')) ? Scalar\String_::KIND_SINGLE_QUOTED : Scalar\String_::KIND_DOUBLE_QUOTED);
             $this->semValue = new Scalar\String_(Scalar\String_::parse($this->semStack[$stackPos-(1-1)]), $attrs);
+            $this->semValue->rawString = $this->semStack[$stackPos-(1-1)];
```
localheinz: What do you think about adjusting the indentation here?

Looks like both lines are incorrectly indented.

Billz95:
Thanks for the reviews, I've updated my code accordingly and squashed them to one commit :sunny:

othercorey:
@nikic Is this something you plan to expect to approve?

TomasVotruba:
Similar node would be useful for `LNumber` and `DNumber`. Now it's impossible to separate `1_000` and `1000` and we have to work with tokens in Rector.

TomasVotruba:
@nikic If you plan to accept this, I can submit PR with rebasd and correct implementation in grammer recipe.

othercorey:
@TomasVotruba Might be better to just create a new PR. He doesn't seem to respond to some PRs.

nikic:
Closing in favor of #831, which implements this as an attribute.

PR #587, opened 2019-02-24
Create . gitattributes

carusogabriel:
`.gitattributes` allows us to configure folders and files that won't be installed when one call `--prefer-dist` in Composer.

An article about it: https://madewithlove.be/gitattributes/.


nikic:
See #178...

PR #596, opened 2019-04-09
AA-0002 fixed the Call to undefined method PhpParser\Node\Expr\Variable::toString() issue

ahmed-abid:
got this issue :
[Error] Call to undefined method PhpParser\Node\Expr\Variable::toString()

while running some tests using AspectMock caused by this line:

https://github.com/goaop/framework/blob/72521d026692131f4052c1d7b3dad15c846cdd0f/src/Instrument/Transformer/MagicConstantTransformer.php#L94

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,3 @@ vendor/
 composer.lock
 grammar/kmyacc.exe
 grammar/y.output
-.idea/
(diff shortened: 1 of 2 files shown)


nikic:
I don't think this is the right way to resolve the issue. Instead the linked code should check for `$methodCallNode->name instanceof Identifier` first.

Adding `toString()` here makes your code work without error, but do something you likely didn't intend: It will accept not just `$foo->getFileName()`, but also `$foo->$getFileName()`.

nikic:
Closing per previous comment.

PR #601, opened 2019-04-19
Fix assert warning on files with no code, only 1 comment

### LINKED ISSUE #589: Warning message on PHP files consisting of only comments
I'm using [`rectorphp/rector`](https://github.com/rectorphp/rector) and when I run the following command, I get a warning message when it found a PHP file where the contents were only comments.

```
$ ./vendor/bin/rector process app --level laravel58
```

Contents of *`app/Test.php`*:

```php
<?php

// A single-line comment
```

Output:

```
PHP Warning:  assert(): assert($itemStartPos >= 0 && $itemEndPos >= 0) failed in
vendor/nikic/php-parser/lib/PhpParser/PrettyPrinterAbstract.php on line 753
```

fitztrev:
This is my attempt to fix #589, when a file is encountered with no PHP code, but does have a comment. Example:

```php
<?php

// a comment
```

This is my first look into this project's codebase so I'm not very familiar with it. Please review and let me know if this solution would work or not for any reason.

### DIFF
--- a/lib/PhpParser/PrettyPrinterAbstract.php
+++ b/lib/PhpParser/PrettyPrinterAbstract.php
@@ -750,7 +750,10 @@ abstract class PrettyPrinterAbstract
 
                 $itemStartPos = $origArrItem->getStartTokenPos();
                 $itemEndPos = $origArrItem->getEndTokenPos();
-                \assert($itemStartPos >= 0 && $itemEndPos >= 0);
+
+                if ($origArrItem instanceof Stmt\Nop && $itemEndPos > -1) {
+                    \assert($itemStartPos >= 0 && $itemEndPos >= 0);
+                }
 
                 if ($itemEndPos < $itemStartPos) {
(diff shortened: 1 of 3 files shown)


fitztrev:
I'm not sure if the Travis 7.4snapshot build is failing because of this PR or an unrelated issue.

fitztrev:
@nikic does this seem OK?

nikic:
Sorry for the delay ... while this avoids the assertion, I think it only hides the issue. We shouldn't end up without end attributes in this case, it probably needs some kind of fix in the parser implementation.

fitztrev:
Ok, thanks for looking. I'm not familiar with the parser implementation so I'm unable to fix.

nikic:
I've implemented a fix for the attribute assignment in the parser in https://github.com/nikic/PHP-Parser/commit/60d025a914aff02d9d7403bbe858e2553fa61f24.

PR #602, opened 2019-04-21
[PHP 7.4] Add arrow function

TomasVotruba:
This PR based on PR to php-src: https://github.com/php/php-src/pull/3941

Current [RFC voting](https://wiki.php.net/rfc/arrow_functions_v2) is 33:5, so if nothing complicated happens in implementation, this feature will get accepted to PHP 7.4.

I need feedback on this. I'm not sure the `ArrowFunction` should inherit `FunctionLike` - because `ArrowFunction` has exactly 1 statement, not more.

1. Should new interface be created?
2. Should `FunctionLike` be parrent to the new interface?
3. Should it extend `Closure`?

I tried to mimic php-src grammar [`Zend/zend_language_parser.y`](https://github.com/php/php-src/pull/3941/files#diff-7eff82c2c5b45db512a9dc49fb990bb8R996)


grammar/php7.y
```diff
@@ -729,6 +729,9 @@ expr:
     | T_YIELD expr                                          { $$ = Expr\Yield_[$2, null]; }
     | T_YIELD expr T_DOUBLE_ARROW expr                      { $$ = Expr\Yield_[$4, $2]; }
     | T_YIELD_FROM expr                                     { $$ = Expr\YieldFrom[$2]; }
+    // used from original php-src PR: https://github.com/php/php-src/pull/3941/files#diff-7eff82c2c5b45db512a9dc49fb990bb8R1002
+    | T_ARROW_FUNCTION optional_ref '(' parameter_list ')' optional_return_type T_DOUBLE_ARROW expr
```
nikic: Missing T_STATIC handling.

lib/PhpParser/Lexer/Emulative.php
```diff
@@ -163,9 +179,28 @@ private function isEmulationNeeded(string $code): bool
...
+    private function processFn(string $code)
+    {
+        if ($this->isFnEmulationNeeded($code) === false) {
+            return;
+        }
+
+        // We need to manually iterate and manage a count because we'll change
+        // the tokens array on the way
+        for ($i = 0, $c = count($this->tokens); $i < $c; ++$i) {
+            if ($this->tokens[$i][0] === T_STRING && $this->tokens[$i][1] === 'fn') {
```
nikic: Need to skip `T_STRING` after `T_OBJECT_OPERATOR`, as it's not interpreted as a keyword there.

lib/PhpParser/Node/Expr/ArrowFunction.php
```diff
@@ -0,0 +1,70 @@
...
+
+    public function getExpr() : Node\Stmt\Expression
+    {
+        return $this->expr;
+    }
+
+    // @todo required by interface, but not really needed
+    /** @return Node\Stmt[] */
+    public function getStmts() : array {
+        return [$this->expr];
```
nikic: `[new Stmt\Return_($this->expr)]` would be more correct here.

lib/PhpParser/PrettyPrinter/Standard.php
```diff
@@ -582,6 +582,11 @@ protected function pExpr_Closure(Expr\Closure $node) {
              . ' {' . $this->pStmts($node->stmts) . $this->nl . '}';
     }
 
+    protected function pExpr_ArrowFunction(Expr\ArrowFunction $node) {
+        return '(' . $this->pCommaSeparated($node->params) . ') => '
```
nikic: Missing `fn` prefix and return type handling.

TomasVotruba:
Thank you! Ready for 2nd feedback

TomasVotruba:
On running parser test:

```php
<?php
fn($a) => $a;
```

with:

```
vendor/bin/phpunit test/PhpParser/CodeParsingTest.php
```

I get error:

```
Syntax error, unexpected T_DOUBLE_ARROW from 2:8 to 2:9
```

The grammar file is somehow wrong, not sure how.

TomasVotruba:
It might be related to this:

- https://github.com/php/php-src/pull/3941/files#diff-93ad74868f98ff7232ebea00007c8b7f

I have a feeling somethign like that is missing in:

- https://github.com/nikic/PHP-Parser/pull/602/files#diff-1cb3f829ad2a9c9fa78bbfb85caaad36

Also there is no information about `T_ARROW_FUNCTION` and maybe should be?

lib/PhpParser/Lexer/TokenEmulator/FnTokenEmulator.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        // We need to manually iterate and manage a count because we'll change
+        // the tokens array on the way
+        for ($i = 0, $c = count($tokens); $i < $c; ++$i) {
+            if ($tokens[$i][0] === T_STRING && $tokens[$i][1] === 'fn') {
+                if (isset($tokens[$i - 1]) && $tokens[$i - 1][0] === T_OBJECT_OPERATOR) {
+                    continue;
+                }
+
+                $tokens[$i][0] = Tokens::T_FN;
```
nikic: This needs to be `self::T_FN`.

TomasVotruba: That failed me in another test, so I changed it. But for parser it works, thanks. I'll continue tomorrow

test/code/formatPreservation/arrow_function.test
```diff
@@ -0,0 +1,21 @@
...
+=>
+$b;
+-----
+<?php
+fn($a) => $a;
+-----
+$stmts[0]->expr->params[] = new Node\Param(new Expr\Variable('b'));
+-----
+<?php
+fn($a, $b) => $a;
```
TomasVotruba: This test fails:

```diff
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
 '<?php
-fn($a, $b) => $a;'
+fn($a, $b) =>
+    $a;'
```

lib/PhpParser/PrettyPrinter/Standard.php
```diff
@@ -582,6 +582,10 @@ protected function pExpr_Closure(Expr\Closure $node) {
              . ' {' . $this->pStmts($node->stmts) . $this->nl . '}';
     }
 
+    protected function pExpr_ArrowFunction(Expr\ArrowFunction $node) {
+        return 'fn(' . $this->pCommaSeparated($node->params) . ') => ' . $this->pStmts([$node->expr]);
```
nikic: Should be just `$this->p($node->expr)`. This is also missing handling for static, by-ref return and return types. You'll probably want to start with a test for just pretty printing, before moving on to the formatting preservation.

TomasVotruba: Good point, will do

test/code/formatPreservation/arrow_function.test
TomasVotruba: Works now :+1:

TomasVotruba:
Tests pass. PHP 7.2 failure is probably related to `fn` in old code of php-src, PHP 7.4 failure is unreleated.

What else should we test?

nikic:
https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/PrettyPrinterAbstract.php#L1198, https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/PrettyPrinterAbstract.php#L1234 and https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/PrettyPrinterAbstract.php#L1277 need to be added for arrow functions with corresponding tests: Adding and removing return type & adding and removing params. You already have a test for params, but because it uses the default formatting, it does not show whether formatting is preserved.

test/code/formatPreservation/arrow_function.test
```diff
@@ -25,6 +25,20 @@ $a, $b
...
+$a;
+-----
+$stmts[0]->expr->params = [];
+-----
+<?php
+fn()
+=>
+$a;
+-----
+<?php
```
TomasVotruba:
```diff
--- Expected
+++ Actual
@@ @@
 '<?php
-fn()
-=>
-$a;'
+fn() => $a;'
```

nikic: Yes, removing first element currently doesn't work. You can add the actual result here with a TODO comment.

nikic: Same for the static & byRef, this is currently not supported.

grammar/tokens.y
```diff
@@ -64,6 +65,7 @@
 %token T_CONTINUE
 %token T_GOTO
 %token T_FUNCTION
+%token T_FN
```
nikic: T_FN also needs to be added in the reserved_non_modifiers list (this is a bug in the php-src implementation as well, I'll fix it there.).

lib/PhpParser/Node/Expr/ArrowFunction.php
```diff
@@ -0,0 +1,72 @@
...
+     * @var null|Node\Identifier|Node\Name|Node\NullableType
+     */
+    public $returnType;
+
+    /**
+     * @var Node\Stmt\Expression
+     */
+    public $expr;
+
+    public function __construct(array $subNodes = [], array $attributes = []) {
```
nikic: Should have a doc comment like this here: https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Node/Expr/Closure.php#L24

lib/PhpParser/Node/Expr/ArrowFunction.php
```diff
@@ -0,0 +1,72 @@
...
+
+use PhpParser\Node;
+use PhpParser\Node\Expr;
+use PhpParser\Node\FunctionLike;
+
+class ArrowFunction extends Expr implements FunctionLike
+{
+    /**
+     * @var bool
+     */
```
nikic: `@var` doc comment should use single-line form.

lib/PhpParser/Node/Expr/ArrowFunction.php
```diff
@@ -0,0 +1,72 @@
...
+    public $params = [];
+
+    /**
+     * @var null|Node\Identifier|Node\Name|Node\NullableType
+     */
+    public $returnType;
+
+    /**
+     * @var Node\Stmt\Expression
```
nikic: This should be `Expr`, not `Node\Stmt\Expression`.

lib/PhpParser/PrettyPrinterAbstract.php
```diff
@@ -1196,6 +1196,7 @@ protected function initializeRemovalMap() {
             'Expr_ArrayDimFetch->dim' => $stripBoth,
             'Expr_ArrayItem->key' => $stripDoubleArrow,
             'Expr_Closure->returnType' => $stripColon,
+            'Expr_ArrowFunction->returnType' => $stripColon,
```
nikic: These are sorted alphabetically, so `ArrowFunction` should be before `Closure` (same below).

test/PhpParser/Lexer/EmulativeTest.php
```diff
@@ -108,6 +111,10 @@ public function testErrorAfterEmulation($code) {
 
     public function provideTestLexNewFeatures() {
         return [
+            // PHP 7.4
+            ['fn', [
+                [Tokens::T_FN, 'fn'],
+            ]],
```
nikic: This is not necessary, having it in provideTestReplaceKeywords is enough.

test/PhpParser/NodeVisitor/NameResolverTest.php
```diff
@@ -212,7 +212,7 @@ interface A extends C, D {
     public function a(A $a) : A;
 }
 
-function fn(A $a) : A {}
+function f(A $a) : A {}
 function fn2(array $a) : array {}
```
nikic: Let's rename this one to `f2` as well.

test/code/parser/expr/arrow_function.test
```diff
@@ -0,0 +1,127 @@
+Arrow Functions
+-----
+<?php
```
nikic: Add a parameter type hint somewhere?

nikic: Also `static` is not tested.

grammar/php5.y
```diff
@@ -650,6 +650,13 @@ expr:
     | T_PRINT expr                                          { $$ = Expr\Print_[$2]; }
     | T_YIELD                                               { $$ = Expr\Yield_[null, null]; }
     | T_YIELD_FROM expr                                     { $$ = Expr\YieldFrom[$2]; }
+
+    // used from original php-src PR: https://github.com/php/php-src/pull/3941/files#diff-7eff82c2c5b45db512a9dc49fb990bb8R1002
+    | T_FN optional_ref '(' parameter_list ')' T_DOUBLE_ARROW expr
+          { $$ = Expr\ArrowFunction[['static' => false, 'byRef' => $2, 'params' => $4, 'expr' => $7]]; }
+    | T_STATIC T_FN optional_ref '(' parameter_list ')' T_DOUBLE_ARROW expr
+          { $$ = Expr\ArrowFunction[['static' => true, 'byRef' => $3, 'params' => $5, 'expr' => $8]]; }
```
nikic: In this case it's probably better to not support this in the PHP 5 parser, as the syntax is incomplete (no return type). If T_FN is added to reserved_non_modifiers, then we won't run into the issue with the token being unused in one parser.

lib/PhpParser/Lexer/TokenEmulator/TokenEmulatorInterface.php
```diff
@@ -0,0 +1,17 @@
+<?php declare(strict_types=1);
+
+namespace PhpParser\Lexer\TokenEmulator;
+
+interface TokenEmulatorInterface
```
nikic: I'm not sure I like this abstraction... I think I would prefer to do all emulation in a single loop (apart from the flexible doc strings, which have a different and much more complicated emulation approach).

test/PhpParser/NodeVisitor/NameResolverTest.php
TomasVotruba: Great idea! :+1:

lib/PhpParser/Lexer/TokenEmulator/TokenEmulatorInterface.php
TomasVotruba: I see. The `PhpParser\Lexer\Emulative` class was getting quite big and it was very hard to orient in it. Adding single token took 3 places to add functionality into.

Now all it takes to add a newline in the constructor. Code responsible for the new emulated token is in own class and easier to orientate.

I'll try to remove one loop.

grammar/php5.y
TomasVotruba: Agreed :+1:

lib/PhpParser/Lexer/TokenEmulator/TokenEmulatorInterface.php
TomasVotruba: I can't figure out how to do it better.

Do you have any tips how to do it like you need and keep it separated in own classes?

TomasVotruba:
I resolved all the comments, ready to review

TomasVotruba:
What needs to be done here?

I need this feature so I can add and test migration path in Rector.

nikic:
7.4 and nightly builds are green now. We should update the used php-src archive to make the 7.2 build also pass.

TomasVotruba:
>  We should update the used php-src archive to make the 7.2 build also pass

Is that part of `test_old`? How should I do that?

nikic:
@TomasVotruba Yes. https://github.com/nikic/PHP-Parser/blob/master/test_old/run-php-src.sh needs to be changed to fetch a 7.4 archive instead.

TomasVotruba:
I see, just a sec :+1: 

Recently I've also noticed "The return keyword can be omitted" in https://stitcher.io/blog/short-closures-in-php 

Is that true?

```php
$ids = array_map(fn(Post $post): int => $post->id, $posts);
$ids = array_map(fn(Post $post): return int => $post->id, $posts);
```

If so, I shall add test case for that

nikic:
> Is that true?

No, adding a return is not allowed there.

TomasVotruba:
I though so

lib/PhpParser/Lexer/TokenEmulator/FnTokenEmulator.php
```diff
@@ -0,0 +1,66 @@
...
+            return false;
+        }
+
+        return strpos($code, 'fn') !== false;
+    }
+
+    public function emulate(string $code, array $tokens) : array
+    {
+        // We need to manually iterate and manage a count because we'll change
+        // the tokens array on the way
```
nikic: Using a foreach would actually be fine in this instance.

lib/PhpParser/Lexer/TokenEmulator/FnTokenEmulator.php
```diff
@@ -0,0 +1,66 @@
...
+        return $tokens;
+    }
+
+    /**
+     * @param mixed[] $tokens
+     * @return mixed[]|null
+     */
+    private function getPreviousNonSpaceToken(array $tokens, int $start)
+    {
+        for ($i = $start - 1; $i > 0; --$i) {
```
nikic: Shouldn't this be `>= 0`? Not that it makes a difference in the context where this is used...

lib/PhpParser/Lexer/TokenEmulator/TokenEmulatorInterface.php
```diff
@@ -0,0 +1,17 @@
...
+{
+    public function getTokenId(): int;
+
+    public function getParserTokenId(): int;
+
+    public function isEmulationNeeded(string $code) : bool;
```
nikic: nit: The placement of `:` is inconsistent between the first two and last two methods. (I don't have a preference either way -- is there a standard for this?)

lib/PhpParser/Node/Expr/ArrowFunction.php
```diff
@@ -0,0 +1,70 @@
...
+    /** @var Expr */
+    public $expr;
+
+    /**
+     * @param array $subNodes   Array of the following optional subnodes:
+     *                          'static'     => false   : Whether the closure is static
+     *                          'byRef'      => false   : Whether to return by reference
+     *                          'params'     => array() : Parameters
+     *                          'returnType' => null    : Return type
+     *                          'expr'       => Expr    : Expression body
```
nikic: Missing `$attributes`?

lib/PhpParser/PrettyPrinterAbstract.php
```diff
@@ -1232,6 +1233,7 @@ protected function initializeInsertionMap() {
             'Expr_ArrayDimFetch->dim' => ['[', false, null, null],
             'Expr_ArrayItem->key' => [null, false, null, ' => '],
             'Expr_Closure->returnType' => [')', false, ' : ', null],
+            'Expr_ArrowFunction->returnType' => [')', false, ' : ', null],
```
nikic: Should go above closure for alphabetical order.

test/code/parser/expr/arrow_function_with_return_type.test
```diff
@@ -0,0 +1,22 @@
+Arrow Functions with Return Types
```
nikic: Now that it's PHP 7 only, I think this file can be merged into the main arrow_function.test.

lib/PhpParser/Lexer/TokenEmulator/FnTokenEmulator.php
TomasVotruba: You're right

lib/PhpParser/Lexer/TokenEmulator/TokenEmulatorInterface.php
TomasVotruba: Most coding standard tools prefer `(): int`

- https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/37cc9c62f71a4cd49d9427aea729954f0ca35fe2/tests/Fixer/FunctionNotation/ReturnTypeDeclarationFixerTest.php#L81-L132
- https://github.com/squizlabs/PHP_CodeSniffer/pull/1219/files#diff-255eee6b30a184bc223a3ebd60e79fe2

So do I :)

---

I think there was some issue about the space, that's why I kept it this way. I'll make it consistent

TomasVotruba:
I was trying to rebase pull-request to merge the fixup! commits, but I keep failing on so many conflicts.

Could you merge it locally or squash to one commit on merge?

nikic:
Merged as squash. Thanks for your work on this!

Next up: https://wiki.php.net/rfc/spread_operator_for_array has been accepted :)

TomasVotruba:
> Merged as squash. Thanks for your work on this!

Awesome :+1: 

> Next up: wiki.php.net/rfc/spread_operator_for_array has been accepted :)

Gimme a sec :laughing:

PR #603, opened 2019-04-23
Enable PHP zend.assertions on Travis builds

fitztrev:
This project makes use of PHP's `assert()`, so this enables the php.ini directive during Travis builds.

### DIFF
--- a/travis.php.ini
+++ b/travis.php.ini
@@ -1,0 +1,1 @@
+zend.assertions = 1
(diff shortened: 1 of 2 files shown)


travis.php.ini
```diff
@@ -0,0 +1 @@
+zend.assertions = 1
```
nikic: Isn't this already the default?

fitztrev: Yes, you're right. I just tested. When I PRed rectorphp/rector#1284 I thought I was only able to see the Warning if I added this.

Issue #605, opened 2019-04-28
Request: produce clean bin/php-parse output on stdout


cyrus-and:
That is, anything that is not the *actual output* should go to stderr. This would allow for example to do:

```console
$ vendor/bin/php-parse -j test.php >test.json
```

Now instead `test.json` begins with:

```
====> File test.php:
==> JSON dump:
```

This is a trivial change but I'm not sure if you're interested. Also trivial would be to write a custom script, but since there's `bin/php-parse` already, it would be nice to just use it.

nikic:
Sound good to me :)

PR #606, opened 2019-04-28
Print messages to stderr in bin/php-parse and fix exit status

cyrus-and:
This is a patch for #605, I have no knowledge of the internals of this library so I touched only `bin/php-parse`; it should be enough.


nikic:
Thanks!

PR #615, opened 2019-06-15
[PHP 7.4] add numeric literal separator to EmulativeLexer

### LINKED ISSUE #614: [PHP 7.4] Add Numeric Literal Separator
- add support to `EmulativeLexer`
- RFC: https://wiki.php.net/rfc/numeric_literal_separator
- PHP src PR: https://github.com/php/php-src/pull/4165

Assign → @TomasVotruba

TomasVotruba:
Closes #614 

I need feedback, not sure how to continue now


lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorTokenEmulator.php
```diff
@@ -0,0 +1,65 @@
...
+        return null;
+    }
+
+    /**
+     * @return int|null
+     */
+    public function getParserTokenId()
+    {
+        return null;
+    }
```
nikic: These should probably be dropped from the interface, and token registration moved back into the emulative lexer class only -- splitting this into separate emulators doesn't really make a lot of sense due to the uniqueness requirement on token IDs.

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorTokenEmulator.php
```diff
@@ -0,0 +1,65 @@
...
+
+    public function isEmulationNeeded(string $code) : bool
+    {
+        // skip version where this is supported
+        if (version_compare(\PHP_VERSION, Emulative::PHP_7_4, '>=')) {
+            return false;
+        }
+
+        return strpos($code, '_') !== false;
```
nikic: This is going to match pretty much any code. We should at least limit this to `[0-9a-fA-F]_[0-9a-fA-F]`, though that will still match lots of code.

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorTokenEmulator.php
```diff
@@ -0,0 +1,65 @@
...
+        // We need to manually iterate and manage a count because we'll change
+        // the tokens array on the way
+        $line = 1;
+        for ($i = 0, $c = count($tokens); $i < $c; ++$i) {
+            if (!isset($tokens[$i + 1])) {
+                continue;
+            }
+
+            $nextToken = $tokens[$i + 1];
+            if (in_array($tokens[$i][0], [T_LNUMBER, T_DNUMBER], true) && $nextToken[0] === T_STRING && strpos($nextToken[1], '_') === 0) {
```
nikic:
```suggestion
            if (($tokens[$i][0] === T_LNUMBER || $tokens[$i][0] === T_DNUMBER) && $nextToken[0] === T_STRING && $nextToken[1][0] === '_') {
```

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorTokenEmulator.php
```diff
@@ -0,0 +1,65 @@
...
+                $numericVault = $tokens[$i][1] . ltrim($nextToken[1], '_');
```
nikic: This should retain the `_` in the token -- they need to be stripped when parsing the number instead. In here for example: https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Node/Scalar/LNumber.php#L43

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorTokenEmulator.php
```diff
@@ -0,0 +1,65 @@
...
+
+                // merge this and next token
+                array_splice($tokens, $i, 2, [
+                    [$tokens[$i][0], $numericVault, $line]
```
nikic: This may end up using the wrong type. For example `1_0.0` should become a DNUMBER, not an LNUMBER.

nikic: Additionally, it may be necessary to handle more than two tokens :( For example `1_0e+10` will be tokenized as `LNUMBER(1) STRING(_0e) '+' LNUMBER(10)`.

This is going to get ugly...

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorTokenEmulator.php
TomasVotruba:
```php
$nextToken[1][0]
```

What about string  `''`?

TomasVotruba:
Thanks for review!

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorTokenEmulator.php
TomasVotruba: I see

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorTokenEmulator.php
TomasVotruba: Why the `in_array` → `===` change? Performance or something else?

TomasVotruba:
Ready for second review.

Boarding :plane: :), will check after 22

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorTokenEmulator.php
TomasVotruba: Agreed

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorTokenEmulator.php
TomasVotruba: Horay :)

test/PhpParser/Lexer/EmulativeTest.php
```diff
@@ -121,31 +124,50 @@ public function testErrorAfterEmulation($code) {
...
+            ['1_000_000_000.0', [
+                [Tokens::T_DNUMBER, '1_000_000_000.0']
+            ]],
+            ['1_0e+10', [
+                [Tokens::T_LNUMBER, '1_0e+10']
+            ]],
+            // @todo how to detect this is float? strlen without "_"?
+            ['0b1011010101001010_110101010010_10101101010101_0101101011001_110111100', [
+                [Tokens::T_DNUMBER, '0b1011010101001010_110101010010_10101101010101_0101101011001_110111100'],
+            ]]
```
TomasVotruba: Need help with this one, how do we know it's float?

nikic: I believe we need to use bindec() and check if the result is a float (same for other bases).

TomasVotruba: Thanks

TomasVotruba:
```diff
<?php

$value = 1_100;

-----
!!php7
array(
    0: Stmt_Expression(
        expr: Expr_Assign(
            var: Expr_Variable(
                name: value
            )
            expr: Scalar_LNumber(
-                value: 1100
            )
        )
    )
)
```

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorEmulator.php
```diff
@@ -0,0 +1,118 @@
...
+
+    private function resolveIntegerOrFloatToken(string $numericValue): int
+    {
+        $numericValueWithoutUnderscores = str_replace('_', '', $numericValue);
+
+        if (strpos($numericValueWithoutUnderscores, '.') !== false) {
+            return T_DNUMBER;
+        }
+
+        $decimalForm = bindec($numericValueWithoutUnderscores);
```
nikic: We'll need to do the same for the other bases as well (hex, octal, decimal).

test/PhpParser/Lexer/EmulativeTest.php
```diff
@@ -121,6 +124,44 @@ public function testErrorAfterEmulation($code) {
...
+            ]],
+            // more complex
+            ['1_0.0', [
+                [Tokens::T_DNUMBER, '1_0.0']
+            ]],
+            ['1_000_000_000.0', [
+                [Tokens::T_DNUMBER, '1_000_000_000.0']
+            ]],
+            ['1_0e+10', [
+                [Tokens::T_LNUMBER, '1_0e+10']
```
nikic: This should be T_DNUMBER.

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorEmulator.php
```diff
@@ -0,0 +1,118 @@
...
+                return true;
+            }
+
+            if ($token[0] === T_DNUMBER) {
+                return true;
+            }
+
+        } else {
+            // matches cases like "1_0e+10" - @todo actually skips first token, because it's a string
+            if ($token === '+') {
```
nikic: May also be `-`.

test/code/parser/scalar/int_php74.test
```diff
@@ -0,0 +1,20 @@
+Different integer syntaxes
+-----
+<?php
+
+$value = 1_100;
+
+-----
+!!php7
```
nikic: Can drop the `!!php7` (we'll be emulating for any version). Not sure why this test fails with a syntax error though...

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorEmulator.php
TomasVotruba: What exactly do you mean? I never worked with these numbers

nikic: For example `0xFFFF_FFFF_FFFF_FFFF` should also be detected as double.

TomasVotruba: And how can we detect it? With bindec and is_int()?
Code snippet would be better

nikic: If prefix is 0b/0B use bindec, if 0x/0X use hexdec if 0 use octdec (each with is_int test). Otherwise (decimal) use `is_int(+$value)`.

TomasVotruba: Thanks. I need help with one last case.

`10e-10` is now interprested as `T_LNUMBER`

nikic: Doesn't the `+$value` trick work for that?

TomasVotruba: It doesn't seem so https://3v4l.org/mAX0l

nikic:
Aren't those results exactly what we want? Both of those should be DNUMBER.

TomasVotruba:
I see. I thought `10000` is int, not really.

```php
var_dump(10e+3); // float(10000);
```

Allright then :+1:

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorEmulator.php
```diff
@@ -0,0 +1,127 @@
+<?php declare(strict_types=1);
+
+namespace PhpParser\Lexer\TokenEmulator;
+
+use PhpParser\Lexer\Emulative;
+use PhpParser\Parser\Tokens;
+
+final class NumericLiteralSeparatorEmulator
```
nikic: After a rebase, this can implement `TokenEmulator` again.

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorEmulator.php
```diff
@@ -0,0 +1,127 @@
...
+        // the tokens array on the way
+        $line = 1;
+        for ($i = 0, $c = count($tokens); $i < $c; ++$i) {
+            if (!isset($tokens[$i + 1])) {
+                continue;
+            }
+
+            $nextToken = $tokens[$i + 1];
+            if (
+            (in_array($tokens[$i][0], [T_LNUMBER, T_DNUMBER], true) && $nextToken[0] === T_STRING && strpos($nextToken[1], '_') === 0)
```
nikic: For performance reasons, `in_array` should not be used. As there are only two tokens here, they can be checked explicitly.

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorEmulator.php
```diff
@@ -0,0 +1,127 @@
...
+            || ($tokens[$i][0] === T_STRING && in_array($nextToken[0], [T_LNUMBER, T_DNUMBER], true))
```
nikic: In which case will the first token be T_STRING?

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorEmulator.php
```diff
@@ -0,0 +1,127 @@
...
+    {
+        // We need to manually iterate and manage a count because we'll change
+        // the tokens array on the way
+        $line = 1;
+        for ($i = 0, $c = count($tokens); $i < $c; ++$i) {
+            if (!isset($tokens[$i + 1])) {
+                continue;
+            }
+
+            $nextToken = $tokens[$i + 1];
```
nikic: Could also `$token = $tokens[$i]` here, as it's used many times below.

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorEmulator.php
```diff
@@ -0,0 +1,127 @@
...
+
+            if ($token[0] === T_DNUMBER) {
+                return true;
+            }
+
+        } else {
+            // matches cases like "1_0e+10" - @todo actually skips first token, because it's a string
+            if ($token === '+' || $token === '-') {
+                if ($previousToken[0] === T_STRING) {
+                    if ($previousToken[1][strlen($previousToken[1]) - 1] === '_') {
```
nikic: I don't understand this check. Why is it checking for a `_` at the end?

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorEmulator.php
```diff
@@ -0,0 +1,127 @@
...
+        if (strpos($numericValueWithoutUnderscores, '.') !== false) {
+            return T_DNUMBER;
+        }
+
+        if (stripos($numericValueWithoutUnderscores, '0b') === 0) {
+            $decimalForm = bindec($numericValueWithoutUnderscores);
+        } elseif (stripos($numericValueWithoutUnderscores, '0x') === 0) {
+            $decimalForm = hexdec($numericValueWithoutUnderscores);
+        } elseif (stripos($numericValueWithoutUnderscores, '0') === 0) {
+            $decimalForm = octdec($numericValueWithoutUnderscores);
```
nikic: Please add a test case for `0e1_0`, I think it might be handled incorrectly.

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorEmulator.php
nikic: Maybe it's supposed to check for e/E at the end?

Something that probably may go wrong right now is something like `1_000+1`, which should not become a single number...

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorEmulator.php
TomasVotruba: Not sure, I think that was my assumption so it works both way.
I can remove it

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorEmulator.php
TomasVotruba: It is. How to fix it?

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorEmulator.php
TomasVotruba: I'm not sure anymore, probably a guess.

Can be removed with no effect

lib/PhpParser/Lexer/TokenEmulator/NumericLiteralSeparatorEmulator.php
nikic: Maybe an additional `ctype_digit()` check on the string? With that, you can also drop the check for `.` at the start, it should fall out naturally.

nikic:
This turned out to be a lot more complicated than I expected ... what I ended up doing is to match a number at the position and then merging everything that was matched, while possibly splitting of a partial token at the end.

TomasVotruba:
I see, thanks for finishing

PR #617, opened 2019-06-28
Add tolerance for multiple commas in function arguments - closes #616

### LINKED ISSUE #616: Add error nodes for additional missing params?
Given the following

```php
function bar(object $o) : void {
    $o->foo(
      $one,
      ,
    );
}
```

there's no `MethodCall` node - I wonder whether it'd be possible to preserve the node, and allow its args array to contain errors as well as arg objects?

iluuu1994:
- [ ] Does the same need to be added to php5.y?


iluuu1994:
Not sure if the message "A trailing comma is not allowed here" is fitting with no function arguments but I didn't want to change the error message.

iluuu1994:
Thanks for the quick response!

> assuming that it's something like this:

Exactly. I'm implementing LSP `SignatureHelp` support for Psalm. It displays a popup with the method parameters and highlights the current one.

> but what you likely want is to have an Expr\Error node in there

Right now, [I have to do quite a bit of ugly tokenizing](https://github.com/iluuu1994/psalm/commit/f131fa654c4165676998ea5d48f9ad47d1f37428#diff-11f0b4d80267531254d3fdd0fbc67248R1672) to find the opening paren `(` and the commas `,` between arguments. This is because `( $arg1 , | $arg2 )` also belongs to the first argument but there's no way of knowing through PHP-Parsers `startFilePos` and `endFilePos` attributes since the symbol and whitespace tokens are thrown away.

So Psalm won't profit from the additional `Expr\Error` nodes in the AST but I can see how others would. Let me know if I should add this.

nikic:
> So Psalm won't profit from the additional Expr\Error nodes in the AST but I can see how others would. Let me know if I should add this.

I think it would make sense to do so, and probably also make the implementation simpler. Just adding `/* empty */` in `argument` should do it.

nikic:
> Right now, I have to do quite a bit of ugly tokenizing to find the opening paren ( and the commas , between arguments. This is because ( $arg1 | , $arg2 ) also belongs to the first argument but there's no way of knowing through PHP-Parsers startFilePos and endFilePos attributes since the symbol and whitespace tokens are thrown away.

The usual way to do this would be to use getTokenStartPos() / getTokenEndPos() and the already existing token stream. I'm assuming that this is not possible in psalm because it disables those attributes for memory usage reasons?

iluuu1994:
> Just adding /* empty */ creating Expr\Error in argument should do it.

This causes a few conflicts:

```yacc
argument:
      expr            { $$ = Node\Arg[$1, false, false]; }
    | '&' variable    { $$ = Node\Arg[$2, true, false]; }
    | T_ELLIPSIS expr { $$ = Node\Arg[$2, false, true]; }
    | /* empty */     { $$ = Node\Arg[Expr\Error[], false, false]; $this->errorState = 2; }
;
```

> there are 3 shift/reduce and 1 reduce/reduce conflicts

Unfortunately I don't understand yacc well enough to know how to solve this. Any suggestions?

PR #622, opened 2019-07-23
Optimize production build by ignoring dev files

kylekatarnls:


carusogabriel:
@kylekatarnls See #178 and the linked PRs.

kylekatarnls:
All those PRs show that the need of having a clean production package is more required than the need of having doc and tests in every vendor directories of the world. People who want it can use `--prefer-source`, please let users decide, don't force them to download files they don't use.

carusogabriel:
Interesting discussion about this topic in Symfony: https://github.com/symfony/symfony/pull/33579

nikic:
Okay, you win :)

PR #623, opened 2019-07-24
Improve printer compatibility with PSR-2/12

SerafimArts:
## Fixed whitespace before Closure "use" statement.

Before:
```php
function ($arg) use($var) { ... }
```

After:
```php
function ($arg) use ($var) { ... }
//                 ^ whitespace added
```

Reason: PSR-2 https://www.php-fig.org/psr/psr-2/#6-closures

## Removed whitespace prefix before typehint colon.

Before:
```php
function example() : void { ... }
```

After:
```php
function example(): void { ... }
//                ^ whitespace removed
```

Reason: PSR-12 https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#45-method-and-function-arguments

### DIFF
--- a/lib/PhpParser/PrettyPrinter/Standard.php
+++ b/lib/PhpParser/PrettyPrinter/Standard.php
@@ -584,8 +584,8 @@ class Standard extends PrettyPrinterAbstract
         return ($node->static ? 'static ' : '')
              . 'function ' . ($node->byRef ? '&' : '')
              . '(' . $this->pCommaSeparated($node->params) . ')'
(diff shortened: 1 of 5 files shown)


nikic:
Generally this looks reasonable, but I don't think I can merge this into 4.x due to the backwards compatibility break this causes.

SerafimArts:
I propose to merge this PR into the future 4.3 version.

Test fails associated with incorrect work on PHP 7.4

nikic:
Variants of these changes have been implemented for the next major in the meantime.

PR #629, opened 2019-08-29
add getProperties() to Class_/Trait_ and getConstants() to Class_/Interface_

TomasVotruba:
As there is `getMethods()` at `ClassLike`, I often tend to use `getProperties()` and `getConstants()`, but there are none. I understood that's probably because only `getMethods()` is in `ClassLike` for all 3 sub-classes.

Yet the API is inconsistent and confusing. This would fix that.

```php
foreach ($class->stmts as $stmt) {
    if (! $stmt instanceof Property) {
        continue;
    }

    // do something
}   
```

vs.

```php
foreach ($class->getMethods() as $method) {
    // do something
}   
```

This keeps happening to me last 2 years so I finally do something about it

### DIFF
--- a/test/PhpParser/Node/Stmt/InterfaceTest.php
+++ b/test/PhpParser/Node/Stmt/InterfaceTest.php
@@ -3,6 +3,7 @@
 namespace PhpParser\Node\Stmt;
 
 use PhpParser\Node;
+use PhpParser\Node\Scalar\String_;
 
 class InterfaceTest extends \PHPUnit\Framework\TestCase
 {
@@ -11,7 +12,7 @@ class InterfaceTest extends \PHPUnit\Framework\TestCase
             new ClassMethod('foo'),
             new ClassMethod('bar'),
         ];
-        $interface = new Class_('Foo', [
+        $interface = new Interface_('Foo', [
             'stmts' => [
                 new Node\Stmt\ClassConst([new Node\Const_('C1', new Node\Scalar\String_('C1'))]),
                 $methods[0],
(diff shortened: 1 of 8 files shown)


test/PhpParser/Node/Stmt/InterfaceTest.php
```diff
@@ -11,7 +12,7 @@ public function testGetMethods() {
             new ClassMethod('foo'),
             new ClassMethod('bar'),
         ];
-        $interface = new Class_('Foo', [
+        $interface = new Interface_('Foo', [
```
TomasVotruba: Fixing test here :)

nikic:
I think these should just be on ClassLike. I know not everything is supported everywhere, but that just means you get an empty array (also, at the parser level they are supported anyway...)

TomasVotruba:
Even if constants are not allowed in traits and properties are not allowed in interfaces?

TomasVotruba:
CI fail is not related to this feature, but rather some bug in 
https://travis-ci.org/nikic/PHP-Parser/jobs/578377375#L346

```bash
The command "if [ $TRAVIS_PHP_VERSION = '7.2' ]; then test_old/run-php-src.sh; fi" exited with 1.
```

nikic:
@TomasVotruba Fixed with https://github.com/nikic/PHP-Parser/commit/5644a916bc92486eb21cdc89f80033f1fbc716d2.

nikic:
> Even if constants are not allowed in traits and properties are not allowed in interfaces?

I think so. For a couple of reasons:

 * I expect that at least properties in interfaces will be supported in the future.
 * While this is forbidden by the language, the parser still accepts both constants in traits and properties in interfaces.
 * For the user, it is easier to just treat this generically -- you can write code that calls `$classLike->getProperties()` and just get back an empty array for interfaces, but don't have to do anything to explicitly distinguish those cases.

TomasVotruba:
Ok, I'd prefer it too, less code to maintain and more ClassLikely :)

Will re-do

TomasVotruba:
Thank you :+1:

PR #633, opened 2019-10-04
README - link to ircmaxell/PHP-Yacc as well

SjonHortensius:


nikic:
Before adding this to the README, we should probably add support for using `php-yacc` in `rebuildParsers.php`.

SjonHortensius:
I do not disagree but it seems to be as simple as updating the `$kmyacc` call, removing the unsupported `-l` options

nikic:
I did the necessary fixes on the php-yacc side to make it a drop-in replacement and switched PHP-Parser to use it by default in https://github.com/nikic/PHP-Parser/commit/eacc5dbe190857ac1f7fa9bc94821f558b35f7ea. `kmyacc` no longer needed.

PR #642, opened 2019-11-17
Add union type to ParamBuilder and BuilderHelpers

TomasVotruba:


lib/PhpParser/Builder/Param.php
```diff
@@ -44,7 +44,7 @@ public function setDefault($value) {
     /**
      * Sets type for the parameter.
      *
-     * @param string|Node\Name|Node\NullableType $type Parameter type
+     * @param string|Node\Name|Node\NullableType|Node\UnionType $type Parameter type
```
nikic: normalizeType() needs to be changed to allow it as well.

TomasVotruba:
Just randomly came here... is anything missing to merge?

lib/PhpParser/BuilderHelpers.php
```diff
@@ -158,9 +158,9 @@ private static function normalizeNameCommon($name, bool $allowExpr) {
      * In particular, builtin types become Identifiers, custom types become Names and nullables
      * are wrapped in NullableType nodes.
      *
-     * @param string|Name|Identifier|NullableType $type The type to normalize
+     * @param string|Name|Identifier|NullableType|UnionType $type The type to normalize
```
nikic: Missing use.

lib/PhpParser/BuilderHelpers.php
```diff
@@ -158,9 +158,9 @@ private static function normalizeNameCommon($name, bool $allowExpr) {
...
      *
-     * @return Name|Identifier|NullableType The normalized type
+     * @return Name|Identifier|NullableType|UnionType The normalized type
      */
     public static function normalizeType($type) {
         if (!is_string($type)) {
```
nikic: UnionType needs to be allowed in the check below.

TomasVotruba:
I'll look on it

TomasVotruba:
> This needs a test.

Where would you test it? 
There is some test:
https://github.com/nikic/PHP-Parser/blob/master/test/code/parser/stmt/function/unionTypes.test

test/PhpParser/Builder/ParamTest.php
```diff
@@ -114,6 +116,12 @@ public function provideTestTypes() {
...
             ['self', new Node\Name('self')],
+            [new Node\Name('Some\Class'), new Node\Name('Some\Class')],
+        ];
+    }
+
+    public function provideTestNullableTypes() {
+        return [
             ['?array', new Node\NullableType(new Node\Identifier('array'))],
             ['?Some\Class', new Node\NullableType(new Node\Name('Some\Class'))],
             [new Node\Name('Some\Class'), new Node\Name('Some\Class')],
```
nikic: This line is duplicated from above.

TomasVotruba: Fixed

nikic:
Thanks!

TomasVotruba:
You're welcome :+1:

PR #644, opened 2019-12-15
Update .travis.yml

andreybolonin:


nikic:
Looks like I merged the same change in #645, sorry about that.

PR #646, opened 2019-12-27
Add `getProperty` to `ClassLike`

ruudk:
This makes it easier to get a property by name from a ClassLike.

### DIFF
--- a/lib/PhpParser/Node/Stmt/ClassLike.php
+++ b/lib/PhpParser/Node/Stmt/ClassLike.php
@@ -54,6 +54,26 @@ abstract class ClassLike extends Node\Stmt
     }
 
     /**
+     * Gets property with the given name defined directly in this class/interface/trait.
+     *
+     * @param string $name Name of the property
+     *
+     * @return Property|null Property node or null if the property does not exist
+     */
+    public function getProperty(string $name) {
+        foreach ($this->stmts as $stmt) {
+            if ($stmt instanceof Property) {
+                foreach ($stmt->props as $prop) {
+                    if ($prop instanceof PropertyProperty && $name === $prop->name->toString()) {
+                        return $stmt;
+                    }
+                }
+            }
+        }
+        return null;
+    }
+
+    /**
      * Gets all methods defined directly in this class/interface/trait
      *
      * @return ClassMethod[]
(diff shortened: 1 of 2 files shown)


lib/PhpParser/Node/Stmt/ClassLike.php
```diff
@@ -53,6 +53,27 @@ public function getProperties() : array {
         return $properties;
     }
 
+    /**
+     * Gets property with the given name defined directly in this class/interface/trait.
+     *
+     * @param string $name Name of the property (compared case-insensitively)
```
nikic: Property names are case-sensitive.

ruudk: Fixed.

ruudk:
Would it be an idea to also add this for `getConstant`?

lib/PhpParser/Node/Stmt/ClassLike.php
```diff
@@ -53,6 +53,26 @@ public function getProperties() : array {
         return $properties;
     }
 
+    /**
+     * Gets property with the given name defined directly in this class/interface/trait.
```
nikic: I think we need to make more clear here that this returns a `Property` which *contains* the desired property, but may also have other properties. So this is sufficient if you want to get the modifiers or property type, but if you need the initializer, you need to fetch the contained `PropertyProperty`.

We should also have a test for the case where the `Property` has multiple properties, and it's sufficient that one of them matches.

ruudk: > I think we need to make more clear here that this returns a Property which contains the desired property, but may also have other properties. So this is sufficient if you want to get the modifiers or property type, but if you need the initializer, you need to fetch the contained PropertyProperty.

What if we also add a `getProperty` method on `Property` that returns the actual PropertyProperty? Together with that, we can improve the description of `ClassLike::getProperty` to mention that method.

> We should also have a test for the case where the Property has multiple properties, and it's sufficient that one of them matches.

Already added, see `fooBarProp` in diff

ruudk:
@nikic Can we merge this PR? Or do you want me to make any adjustments. Let me know, thanks!

nikic:
Sorry for the delay!

TomasVotruba:
Any ETA on release? I'd love to use this in Rector and drop custom method handling it: 

https://github.com/rectorphp/rector/blob/5f4ca314a9c307c28d40551bff1fc1d3ad4d82a7/src/PhpParser/Node/Manipulator/ClassManipulator.php#L213-L231

PR #647, opened 2020-01-13
add psalm types

orklah:
Hi,

Working on https://github.com/Roave/BetterReflection/issues/525, I needed to refine some types in PHP-Parser.

I tried to either keep readable annotations for IDE like PHPStorm or add a @psalm- annotation.

I take the opportunity to thank you for all the good you did to PHP !

### DIFF
--- a/lib/PhpParser/Node/Stmt/ClassConst.php
+++ b/lib/PhpParser/Node/Stmt/ClassConst.php
@@ -10,16 +10,16 @@ class ClassConst extends Node\Stmt
     public $flags;
     /** 
      * @var Node\Const_[] Constant declarations
-     * @psalm-var list<Node\Const>
+     * @psalm-var list<Node\Const_>
      */
     public $consts;
 
     /**
      * Constructs a class const list node.
      *
-     * @param Node\Const_[] $consts     Constant declarations
-     * @param int           $flags      Modifiers
-     * @param array         $attributes Additional attributes
+     * @param list<Node\Const_> $consts     Constant declarations
+     * @param int               $flags      Modifiers
+     * @param array             $attributes Additional attributes
      */
     public function __construct(array $consts, int $flags = 0, array $attributes = []) {
         $this->attributes = $attributes;
--- a/lib/PhpParser/Node/Stmt/ClassConst.php
+++ b/lib/PhpParser/Node/Stmt/ClassConst.php
@@ -8,7 +8,10 @@ class ClassConst extends Node\Stmt
 {
     /** @var int Modifiers */
     public $flags;
-    /** @var Node\Const_[] Constant declarations */
+    /** 
+     * @var Node\Const_[] Constant declarations
+     * @psalm-var list<Node\Const>
+     */
     public $consts;
 
     /**
--- a/lib/PhpParser/PrettyPrinter/Standard.php
+++ b/lib/PhpParser/PrettyPrinter/Standard.php
@@ -449,6 +449,7 @@ class Standard extends PrettyPrinterAbstract
 
     protected function pExpr_Cast_Double(Cast\Double $node) {
         $kind = $node->getAttribute('kind', Cast\Double::KIND_DOUBLE);
+        $cast = '';
         if ($kind === Cast\Double::KIND_DOUBLE) {
             $cast = '(double)';
         } elseif ($kind === Cast\Double::KIND_FLOAT) {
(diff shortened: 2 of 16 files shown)


staabm:
This goes into the same direction like I requested in https://github.com/nikic/PHP-Parser/issues/726

Tools like psam, phpstan, rector the phpstan/psalm ecosystem (namely plugins) etc. Would really benefit from getting types into this lib.

lib/PhpParser/PrettyPrinter/Standard.php
```diff
@@ -449,6 +449,7 @@ protected function pExpr_Cast_Int(Cast\Int_ $node) {
 
     protected function pExpr_Cast_Double(Cast\Double $node) {
         $kind = $node->getAttribute('kind', Cast\Double::KIND_DOUBLE);
+        $cast = '';
```
azjezz, line 452:
```suggestion
        /** @psalm-var Cast\Double::KIND_* $kind */
        $kind = $node->getAttribute('kind', Cast\Double::KIND_DOUBLE);
```

this will remove the issue about possibly undeclared `$cast`, tell psalm that `$cast` is always `non-empty-string`, and also help catch bugs when comparing it to other types :)

orklah: Indeed, I knew way less about Psalm then, I could fix that better now :)

Thanks for the review, but I'm not sure @nikic is interested by improving type documentation in this repo. This PR has waited for months without answers, I kinda made my peace with that!

lib/PhpParser/Node/Stmt/ClassConst.php
```diff
@@ -8,15 +8,18 @@ class ClassConst extends Node\Stmt
...
+     */
     public $consts;
 
     /**
      * Constructs a class const list node.
      *
-     * @param Node\Const_[] $consts     Constant declarations
-     * @param int           $flags      Modifiers
-     * @param array         $attributes Additional attributes
+     * @param list<Node\Const_> $consts     Constant declarations
```
Chemaclass, line 20: I think you need the psalm prefix when using `list`, right?
`@psalm-param list<...`

lib/PhpParser/Node/Stmt/ClassConst.php
```diff
@@ -8,15 +8,18 @@ class ClassConst extends Node\Stmt
...
     public $flags;
-    /** @var Node\Const_[] Constant declarations */
+    /** 
+     * @var Node\Const_[] Constant declarations
+     * @psalm-var list<Node\Const_>
+     */
     public $consts;
 
     /**
      * Constructs a class const list node.
```
Chemaclass, line 18: I think you can remove this comment. What do you think?

lib/PhpParser/Node/Stmt/ClassConst.php
```diff
@@ -8,15 +8,18 @@ class ClassConst extends Node\Stmt
 {
     /** @var int Modifiers */
     public $flags;
-    /** @var Node\Const_[] Constant declarations */
+    /** 
+     * @var Node\Const_[] Constant declarations
+     * @psalm-var list<Node\Const_>
```
Chemaclass, line 13: I think keeping only `@psalm-var list<Node\Const_> Constant declarations` would be sufficient.
There is no need to keep the other one, right?

lib/PhpParser/Node/Stmt/ClassConst.php
Chemaclass: Ah! I thought `list<>` was psalm's syntax, therefore to be able to use it you needed to prefix it. But maybe I am wrong. In that case, It's fine :)

orklah: it is, but it's now recognized by other tools and IDEs (phpstan considers it an alias of `array<int, XXX>` and phpstorm sees it as an alias of XXX[])

orklah:
I'm closing this, it's way too old now and it was very incomplete.

As a Psalm maintainer though, I'd very much like to do a full PR with all the types needed for third party tools to consume. Psalm itself has a lot of dedicated code, configs and stubs in order to fully understand PHP-Parser and it would be easier to maintain them upstream instead of duplicating this work into every third party lib (see https://github.com/phpstan/phpstan-php-parser/issues/12).

If you ever feel like it, please ping me and I'll add those in whatever format you see fit.

PR #655, opened 2020-03-24
[PHP 7.4] Make DNumber use fromString() approach like LNumber, keep original value in 'original_value' attribute

TomasVotruba:
Follow up to #615 

## Reasoning

Currently, all the `LNumber` and `DNumber` nodes contain only value without `_`.
Saying that, it's not possible to know, it the original value was with `_` or without `_`.

That way @rectorphp that converts number values to underscored ones fails: 

```diff
-1000
+1_000

# we don't know there already is `_` :(
-10_00
+1_000
```

We could hack this on `AbstractParser` level - I tried it and it would basically mean replacing many nodes and private methods, which is way too hacky.

This also bring more consise api for `LNumber` and `DNumber` construction.

### DIFF
--- a/grammar/php5.y
+++ b/grammar/php5.y
@@ -782,7 +782,7 @@ ctor_arguments:
 
 common_scalar:
       T_LNUMBER                                             { $$ = $this->parseLNumber($1, attributes(), true); }
-    | T_DNUMBER                                             { $$ = Scalar\DNumber[Scalar\DNumber::parse($1)]; }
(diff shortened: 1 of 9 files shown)


TomasVotruba:
Closing as too hard to change. Let's stay with saint defaults

nikic:
What do you mean by "too hard to change"? As in, if you want to use different formatting?

TomasVotruba:
I thought this is not accepted.

PR #661, opened 2020-05-08
[PHP 8.0] Add attributes v2 support

### LINKED ISSUE #659: [PHP 8] attributes 2 support
This might be fun :). I would like to give it try, if you haven't made it yet.

How could we handle it? 
What apart grammar needs to be done?
We could use similar approach as for docblocks.

Also, how should we name it? There are already, attributes at each node.

Grammar: https://github.com/beberlei/php-src/blob/ea9277b9a416a1bd13217ead0a4c58bad1eaaf85/Zend/zend_language_parser.y

TomasVotruba:
Closes #659

<br>

Note, this comes handy (found out 1 hour too later than needed :facepalm: ):

```bash
php test/updateTests.php
```


TomasVotruba:
It's goes pretty well so far. I just got stuck on simple test case: https://github.com/nikic/PHP-Parser/pull/661/commits/8025300c793e74e2bd0acf87c5184f548a4a823a#diff-39ff209fb56c0dd7ab3dcde01b4c6df1

With:

```
1) PhpParser\CodeParsingTest::testParse with data set "var/www/PHP-Parser/test/code/parser/stmt/function/attributes.test#0" ('Attributes (/var/www/PHP-Pars....test)', '<?php\n\n<<Deprecated>>\nfunc...) {\n}', 'array(\n    0: Stmt_Function(...  )\n)', null)
Attributes (/var/www/PHP-Parser/test/code/parser/stmt/function/attributes.test)
--- Expected
+++ Actual
@@ @@
-'array(
-    0: Stmt_Function(
-        byRef: false
-        name: Identifier(
-            name: a
+'Syntax error, unexpected T_SL from 3:1 to 3:2
+Syntax error, unexpected T_STRING, expecting '(' from 4:10 to 4:10
+Syntax error, unexpected '{' from 4:14 to 4:14
```

I assume the `tokens.yaml` has to be updated somehow, right?

TomasVotruba:
Thanks! It helped

1. Now the node attributes are missing in the Attribute node. Any idea how to get them in `nodeWithPhpAttributes`?

2. Where exactly in the StandardPrinter should be Attribute be printed?

grammar/php7.y
```diff
@@ -75,10 +75,33 @@ optional_comma:
...
+annotated_statement:
+      function_declaration_statement                        { $$ = $1; }
+    | class_declaration_statement                           { $$ = $1; }
+
 top_statement:
       statement                                             { $$ = $1; }
-    | function_declaration_statement                        { $$ = $1; }
-    | class_declaration_statement                           { $$ = $1; }
+    | annotated_statement                                   { $$ = $1; }
+    | attributes annotated_statement                        { nodeWithPhpAttributes($2, $1); }
```
nikic, line 104: Instead of something like `nodeWithPhpAttributes()`, every construct that supports attributes should have an `optional_attributes` at the start.

nikic: Basically what you did for anon functions, but preferably without rule duplication.

grammar/php7.y
```diff
@@ -740,8 +774,12 @@ expr:
     | T_YIELD expr T_DOUBLE_ARROW expr                      { $$ = Expr\Yield_[$4, $2]; }
     | T_YIELD_FROM expr                                     { $$ = Expr\YieldFrom[$2]; }
 
+    | attributes T_FN optional_ref '(' parameter_list ')' optional_return_type T_DOUBLE_ARROW expr
+          { $$ = Expr\ArrowFunction[['static' => false, 'byRef' => $2, 'params' => $4, 'returnType' => $6, 'expr' => $8]]; }
     | T_FN optional_ref '(' parameter_list ')' optional_return_type T_DOUBLE_ARROW expr
           { $$ = Expr\ArrowFunction[['static' => false, 'byRef' => $2, 'params' => $4, 'returnType' => $6, 'expr' => $8]]; }
+    | attribute T_STATIC T_FN optional_ref '(' parameter_list ')' optional_return_type T_DOUBLE_ARROW expr
```
nikic, line 781:
```suggestion
    | attributes T_STATIC T_FN optional_ref '(' parameter_list ')' optional_return_type T_DOUBLE_ARROW expr
```

grammar/php7.y
nikic: If you write the `attributes` as:

```
attributes:
      /* empty */                                           { init(); }
    | attributes attribute                                  { push($1, $2); }
;
```

you can probably get away without an explicit optional_attributes rule.

lib/PhpParser/Node/Stmt/Attribute.php
```diff
@@ -0,0 +1,36 @@
+<?php declare(strict_types=1);
+
+namespace PhpParser\Node\Stmt;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr;
+
+class Attribute extends Node\Stmt
```
nikic, line 8: I'd suggest to extend `NodeAbstract` directly here, because attributes are not really statements.

lib/PhpParser/Node/Stmt/Attribute.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    /**
+     * @var Node\Name
+     */
```
nikic, line 12:
```suggestion
    /** @var Node\Name */
```

etc

lib/PhpParser/Node/Stmt/Attribute.php
```diff
@@ -0,0 +1,36 @@
...
+    public $name;
+    /**
+     * @var Expr[]|null
+     */
+    public $values;
```
nikic, line 17: Maybe it would be better to use an empty array instead of null? I don't think we need to preserve difference between `<<A>>` and `<<A()>>` here. (Similar to how `new A` vs `new A()` is not preserved in AST.)

nikic: Maybe this should actually be `Arg[]`. These are really constructor arguments and I see a potential forward-compatibility issue here, if we introduce named arguments.

lib/PhpParser/NodeVisitor/NameResolver.php
```diff
@@ -137,6 +137,18 @@ public function enterNode(Node $node) {
...
                 }
             }
+        } elseif ($node instanceof Stmt\Function_
+                    || $node instanceof Expr\Closure
+                    || $node instanceof Stmt\ClassMethod
+                    || $node instanceof Expr\ArrowFunction
+                    || $node instanceof Stmt\Property
+                    || $node instanceof Stmt\Const_
+                    || $node instanceof Stmt\Class_
+                    || $node instanceof Node\Param
```
nikic, line 147: Some of these types are already previously checked and this code will not be reached. I would suggest adding a `resolveAttributes()` method and calling it in appropriate places.

grammar/php7.y
TomasVotruba: Good idea! I tried to implement it, hope I understood you correctly.

There is one token conflict but I'm not able to find where :/

TomasVotruba:
I've integreated your comment. It's ready for next review

lib/PhpParser/Node/Stmt/Function_.php
```diff
@@ -31,19 +34,21 @@ class Function_ extends Node\Stmt implements FunctionLike
      *                           'returnType' => null   : Return type
      *                           'stmts'      => array(): Statements
      * @param array  $attributes Additional attributes
+     * @param Attribute[] $phpAttributes
      */
-    public function __construct($name, array $subNodes = [], array $attributes = []) {
+    public function __construct($name, array $subNodes = [], array $attributes = [], array $phpAttributes = []) {
```
nikic, line 39: Instead of an extra argument, this should be extracted from `$subNodes`. Same for other cases accepting `$subNodes`.

nikic:
I pushed a fix for the parser conflicts. Haven't had a chance to review the rest yet.

TomasVotruba:
Thanks, that helped a lot. I'll continue

lib/PhpParser/Node/Stmt/Function_.php
TomasVotruba: Updated.

TomasVotruba:
Tests are now broken with some positions. I really don't understand why or how to fix them. Seems like attributes are not parsed anymore

TomasVotruba:
I tried to rebase on current `master`.

Could you give this priority? New merged PRs makes this hell to maintain :/ 

I think only details are missing, but my knowledge of php grammar errors is not that advanced, so I can fix them.

nikic:
Emulation for new T_ATTRIBUTE token added in https://github.com/nikic/PHP-Parser/commit/f66a32e2df7866d593fa7a1bb582348f772511b0. I've rebased this PR and fixed the shift-reduce conflicts.

Unfortunately the `#[]` syntax also has an attribute grouping feature `#[Attr1, Attr2]`, so I think it will be necessary to change the representation, i.e. introduce an additional intermediary `AttributeGroup` node :/

nikic:
I've mostly finished this up and merged the initial implementation. One thing that is still missing is full FPPP support.

ondrejmirtes:
Awesome! 😍 

But what's FPPP?

Screenshot 2020-09-13 at 21 08 24

(I'm not afraid to ask the stupid questions.)

nikic:
FPPP = Formatting-Preserving Pretty Print. It's a bit long to write out :)

ondrejmirtes:
Oh, of course :)

PR #672, opened 2020-06-22
[PHP 8.0] Add match expressions

### LINKED ISSUE #671: [PHP 8.0] Add match expression v2
- PHP wiki: https://wiki.php.net/rfc/match_expression_v2
- RFC Grammar: https://github.com/php/php-src/pull/5371/files#diff-7eff82c2c5b45db512a9dc49fb990bb8

<br>

Note to myself :)

TomasVotruba:
- Official RFC Wiki: https://wiki.php.net/rfc/match_expression_v2
- Covers: https://github.com/php/php-src/pull/5371
- Closes #671 

```php
// After
$statement = match ($this->lexer->lookahead['type']) {
    Lexer::T_SELECT => $this->SelectStatement(),
    Lexer::T_UPDATE => $this->UpdateStatement(),
    Lexer::T_DELETE => $this->DeleteStatement(),
    default => $this->syntaxError('SELECT, UPDATE or DELETE'),
};
```


TomasVotruba:
Ready for review

lib/PhpParser/Node/Stmt/MatchArm.php
```diff
@@ -0,0 +1,30 @@
...
+namespace PhpParser\Node\Stmt;
+
+use PhpParser\Node;
+
+class MatchArm extends Node\Stmt
+{
+    /** @var null|Node\Expr[] */
+    public $condList;
+    /** @var Node\Expr */
+    public $body;
```
TomasVotruba, line 12: Maybe name this `expr` as in other nodes?

TomasVotruba:
The failing job is related to old tests:
https://travis-ci.org/github/nikic/PHP-Parser/jobs/700969156#L328

This test fixture file is using "match" in function name `function match()`: https://github.com/php/php-src/blob/dad377ba5774bf624a0f111d0eca274053124b87/ext/standard/tests/file/fnmatch_variation.phpt#L81

It was fixed in the RFC PR: https://github.com/php/php-src/pull/5371/files#diff-b78eeef7351e763ed076a5c98d6bec84

image


So not related to this PR.

lib/PhpParser/PrettyPrinterAbstract.php
```diff
@@ -1256,6 +1256,7 @@ protected function initializeInsertionMap() {
             'Expr_Yield->value' => [\T_YIELD, false, ' ', null],
             'Param->type' => [null, false, null, ' '],
             'Param->default' => [null, false, ' = ', null],
+            'Stmt_MatchArm->body' => [null, false, ', ', null],
```
TomasVotruba, line 1259: I thought this is the way to add a comma after specific subnode, but it doesn't work.
Any tips?

grammar/php7.y
```diff
@@ -399,6 +400,34 @@ case_separator:
     | ';'
 ;
 
+match:
+      T_MATCH '(' expr ')' '{' match_arm_list '}'           { $$ = Expr\Match_[$3, $6]; }
+;
+
+match_arm_list:
+      /* empty */                                           { $$ = null; }
```
nikic, line 408:
```suggestion
      /* empty */                                           { $$ = []; }
```

grammar/php7.y
```diff
@@ -399,6 +400,34 @@ case_separator:
...
+;
+
+non_empty_match_arm_list:
+      match_arm                                             { init($1); }
+    | non_empty_match_arm_list ',' match_arm                { push($1, $3); }
+;
+
+match_arm:
+      match_arm_cond_list T_DOUBLE_ARROW match_arm_body     { $$ = Stmt\MatchArm[$1, $3]; }
+    | T_DEFAULT T_DOUBLE_ARROW match_arm_body               { $$ = Stmt\MatchArm[NULL, $3]; }
```
nikic, line 419:
```suggestion
    | T_DEFAULT T_DOUBLE_ARROW match_arm_body               { $$ = Stmt\MatchArm[null, $3]; }
```

grammar/php7.y
```diff
@@ -399,6 +400,34 @@ case_separator:
...
+;
+
+match_arm_cond_list:
```
nikic, line 422: You can use expr_list_forbid_comma` instead.

grammar/php7.y
```diff
@@ -399,6 +400,34 @@ case_separator:
...
+      expr                                                  { init($1); }
+    | match_arm_cond_list ',' expr                          { push($1, $3); }
+;
+
+match_arm_body:
```
nikic, line 427: You can directly use `expr` instead.

lib/PhpParser/Lexer/TokenEmulator/MatchTokenEmulator.php
```diff
@@ -0,0 +1,53 @@
...
+            return false;
+        }
+
+        return strpos($code, 'match') !== false;
+    }
+
+    public function emulate(string $code, array $tokens): array
+    {
+        // We need to manually iterate and manage a count because we'll change
+        // the tokens array on the way
```
nikic, line 22: This comment doesn't make sense here (we don't maintain a manual count).

lib/PhpParser/Lexer/TokenEmulator/MatchTokenEmulator.php
```diff
@@ -0,0 +1,53 @@
...
+        foreach ($tokens as $i => $token) {
+            if ($token[0] === T_STRING && $token[1] === 'match') {
```
nikic, line 24: Tokens are case insensitive, needs a strtolower.

lib/PhpParser/Node/Expr/Match_.php
```diff
@@ -0,0 +1,31 @@
...
+
+use PhpParser\Node;
+use PhpParser\Node\Stmt\MatchArm;
+
+class Match_ extends Node\Expr
+{
+    /** @var Node\Expr */
+    public $cond;
+    /** @var MatchArm[] */
+    public $matchArms;
```
nikic, line 13: Possibly just `$arms`? The match seems redundant here.

lib/PhpParser/Node/Expr/Match_.php
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * @param \PhpParser\Node\Stmt\MatchArm[] $matchArms
```
nikic, line 16:
```suggestion
     * @param MatchArm[] $matchArms
```

lib/PhpParser/Node/Stmt/MatchArm.php
nikic: `body` is probably safer in case support for statements is added in the future.

lib/PhpParser/Node/Stmt/MatchArm.php
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * @param null|Node\Expr[] $condList
+     */
+    public function __construct($condList, Node\Expr $body, array $attributes = []) {
+        $this->condList = $condList;
+        $this->body = $body;
+
+        parent::__construct($attributes);
```
nikic, line 21:
```suggestion
        $this->attributes = $attributes;
```
We don't parent parent constructor for performance reasons.

lib/PhpParser/Node/Stmt/Switch_.php
```diff
@@ -12,8 +12,6 @@ class Switch_ extends Node\Stmt
     public $cases;
 
     /**
-     * Constructs a case node.
```
nikic, line 15: Keep this, it's present everywhere. Makes little sense to drop it in only one case.

lib/PhpParser/PrettyPrinterAbstract.php
nikic: This list only takes effect if body is changed from null to Expr, which can't happen. You need the ListInsertionMap for new match arms / new match conditions.

TomasVotruba: Done

lib/PhpParser/Lexer/TokenEmulator/MatchTokenEmulator.php
TomasVotruba: It's only copy pasted from other emulator classes. Should it be removed everywhere or just here?

lib/PhpParser/Node/Stmt/Switch_.php
TomasVotruba: > Makes little sense to drop it in only one case.

My bad. Intent was to clear comment without any value. Will handle the rest of comments.

TomasVotruba: > Makes little sense to drop it in only one case.

My bad. Intent was to clear comment without any value. Will handle the rest of comments.

TomasVotruba: > Makes little sense to drop it in only one case.

My bad. Intent was to clear comment without any value. Will handle the rest of comments.

lib/PhpParser/PrettyPrinterAbstract.php
```diff
@@ -1326,6 +1326,8 @@ protected function initializeListInsertionMap() {
             'Stmt_TraitUseAdaptation_Precedence->insteadof' => ', ',
             'Stmt_Unset->vars' => ', ',
             'Stmt_Use->uses' => ', ',
+            'Stmt_Arm->condList' => ', ',
+            'Stmt_Match->arms' => ', ',
```
TomasVotruba, line 1330: Is this correct? The comma is still missing in the print :P

TomasVotruba, line 1330: Is this correct? The comma is still missing in the print :/

TomasVotruba:
Updated

TomasVotruba:
Thanks for finishing it!

Do you recall where I messed up the grammar so the comma was missing?

nikic:
@TomasVotruba I think everything in the grammar was right, the problem was was the use of pStmts() instead of pCommaSeparatedMultiline() inside the pretty-printer.

TomasVotruba:
I thought I tried both with same bad result. Well, maybe I overlooked it.

Thanks for letting me know :+1: , I'll be patient to it next time

ondrejmirtes:
Is it expected that Emulative lexer will break on existing PHP 7-level code? I'd say that people use Emulative when they want to parse features from newer PHP versions on an older PHP version.

But this change means that the Emulative lexer will fail with parse error on `interface Match` which is valid on PHP 7. Which is pretty inconvenient as PHPUnit contains this interface.

nikic:
@ondrejmirtes Unfortunately the emulative lexer currently only does emulation in one direction. Emulation in the other direction (and ability to pick a precise PHP version to target) would have to be implemented.

PR #681, opened 2020-07-10
Add ParentConnectingVisitor and NodeConnectingVisitor

sebastianbergmann:


lib/PhpParser/NodeVisitor/NodeConnectingVisitor.php
```diff
@@ -0,0 +1,52 @@
...
+ */
+final class NodeConnectingVisitor extends NodeVisitorAbstract
+{
+    /**
+     * @var Node[]
+     */
+    private $stack;
+
+    /**
+     * @var Node
```
GrahamCampbell, line 25: Could be null? `Node|null`

lib/PhpParser/NodeVisitor/NodeConnectingVisitor.php
```diff
@@ -0,0 +1,52 @@
...
+
+use PhpParser\Node;
+use PhpParser\NodeVisitorAbstract;
+
+/**
+ * Visitor that connects a child node to its parent node
+ * as well as its sibling nodes.
+ *
+ * On the child node, the parent node can be access through
+ * <code>$node->getAttribute('parent')</code>, the previous
```
nikic, line 13: Huh, HTML in doc comments is a thing?

lib/PhpParser/NodeVisitor/NodeConnectingVisitor.php
```diff
@@ -0,0 +1,52 @@
...
+namespace PhpParser\NodeVisitor;
+
+use PhpParser\Node;
+use PhpParser\NodeVisitorAbstract;
+
+/**
+ * Visitor that connects a child node to its parent node
+ * as well as its sibling nodes.
+ *
+ * On the child node, the parent node can be access through
```
nikic, line 12:
```suggestion
 * On the child node, the parent node can be accessed through
```

lib/PhpParser/NodeVisitor/ParentConnectingVisitor.php
```diff
@@ -0,0 +1,41 @@
...
+
+use function array_pop;
+use function count;
+use PhpParser\Node;
+use PhpParser\NodeVisitorAbstract;
+
+/**
+ * Visitor that connects a child node to its parent node.
+ *
+ * On the child node, the parent node can be access through
```
nikic, line 13:
```suggestion
 * On the child node, the parent node can be accessed through
```

test/PhpParser/NodeVisitor/NodeConnectingVisitorTest.php
```diff
@@ -0,0 +1,57 @@
...
+
+        $traverser = new NodeTraverser;
+
+        $traverser->addVisitor(new NodeConnectingVisitor);
+
+        $ast = $traverser->traverse($ast);
+
+        $traverser = new NodeTraverser;
+
+        $visitor = new FindingVisitor(
```
nikic, line 28: An shorter way to write this is `(new NodeFinder)->findFirstInstanceof($ast, Else_::class)`. Similar for the other test.

lib/PhpParser/NodeVisitor/NodeConnectingVisitor.php
sebastianbergmann: A sub-set that contains `<code>` is supported, yes.

test/PhpParser/NodeVisitor/NodeConnectingVisitorTest.php
sebastianbergmann: Thanks!

lib/PhpParser/NodeVisitor/NodeConnectingVisitor.php
GrahamCampbell: Most phpdoc parsers treat phpdoc comments as markdown, which is why the code tag is supported (since html can be written in markdown documents). Lots of repos also use single backticks for code. Both are supported by most phpdoc parsers.

sebastianbergmann:
Thank you, @nikic and @GrahamCampbell.

llaville:
These visitors are really useful especially in class (object) context, when we need to retrieve class (object) name of the method (...) when the standard `NameResolver` cannot solve the situation. 

@nikic My question is : When do you plan a new version that will include these new visitors ?

llaville:
Thanks @nikic for the new version 4.7.0

PR #682, opened 2020-07-13
Fixed branch alias

GrahamCampbell:


nikic:
Maybe just drop this?

GrahamCampbell:
The point of the branch alias is to allow people to install dev version of the package, so it's good to keep it. Alternatively you could just name the branches "correctly", which would avoid the need for aliases at all.

nikic:
Can't people just use dev-master?

GrahamCampbell:
No, that's not a version constraint. That's just a hack which gets the master branch, and breaks, for example if the master branch is renamed, or they have another package which depends on `^4.0`, for example.

PR #686, opened 2020-07-16
Fixed rebuildParsers.php

MikaelMayer:
phpyacc is a sh file and should not be run via the php interpreter, but directly.


nikic:
Is this a Windows specific issue?

MikaelMayer:
I'm definitely on Windows. I'm not sure how it works in other environments.
Thanks for checking.

Issue #679, opened 2020-07-08
[PHP 8.0] Support for static as return type


ondrejmirtes:
RFC: https://wiki.php.net/rfc/static_return_type

The parser cannot currently parse this: https://phpstan.org/r/c33ac255-8b28-4c81-a664-e1bf17e5fb91

/cc @TomasVotruba Could you look into it? Looks like you have many great contributions recently .)

nikic:
Static return type support added in https://github.com/nikic/PHP-Parser/commit/7f398769a09a87f60516cb47ba6bf075dde88661.

ondrejmirtes:
Thank you very much!

PR #689, opened 2020-07-25
Fix up tests

GrahamCampbell:
1. PHPUnit 8 doesn't work on PHP 8.0. PHPUnit 9.3(-dev) is required.
2. The "old test" runner was broken by 69c5d48afd1e631290de1aa9806eab0b38bb66a3. This PR fixes that*.

\* Note that the old tests are still failing due to https://github.com/nikic/PHP-Parser/issues/679, but the old tests fundamentally can't run on PHP 7 anymore because of the fact php-parser doesn't distinguish php 7 from 8 like it did with 5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,16 +26,18 @@ matrix:
       php: nightly
       install:
         - composer update --ignore-platform-req=php --no-progress --prefer-dist
-    - name: PHP 7 Integration Tests
-      php: 7.4
+    - name: PHP 7.3 Code on PHP 8.0 Integration Tests
+      php: nightly
+      install:
+        - composer update --ignore-platform-req=php --no-progress --prefer-dist
       script:
-        - test_old/run-php-src.sh 7
-    - name: PHP 8 Integration Tests
-      php: 7.4
+        - test_old/run-php-src.sh 7.3.21
+    - name: PHP 8.0 Code on PHP 7.0 Integration Tests
+      php: 7.0
       script:
-        - test_old/run-php-src.sh 8
+        - test_old/run-php-src.sh 8.0.0beta1
   allow_failures:
-    - name: PHP 8 Integration Tests
+    - name: PHP 8.0 Code on PHP 7.0 Integration Tests
   fast_finish: true
 
 script: vendor/bin/phpunit
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@ matrix:
     - name: PHP 8.0 Unit Tests
       php: nightly
       install:
-        - composer require phpunit/phpunit:^9.3 --dev --no-update
         - composer update --ignore-platform-req=php --no-progress --prefer-dist
     - name: PHP 7 Integration Tests
       php: 7.4
@@ -36,7 +35,6 @@ matrix:
       script:
         - test_old/run-php-src.sh 8
   allow_failures:
-    - name: PHP 8.0 Unit Tests
     - name: PHP 8 Integration Tests
   fast_finish: true
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,35 +1,45 @@
 language: php
 dist: xenial
 
-cache:
-  directories:
-    - $HOME/.composer/cache
+before_install: composer self-update --2
 
-php:
-  - 7.0
-  - 7.1
-  - 7.2
-  - 7.3
-  - 7.4
-  - nightly
-
-install:
-  - if [ $TRAVIS_PHP_VERSION = '7.0' ]; then composer require satooshi/php-coveralls '~1.0'; fi
-  - |
-    if [ $TRAVIS_PHP_VERSION = 'nightly' ]; then
-      composer install --prefer-dist --ignore-platform-reqs;
-    else
-      composer install --prefer-dist;
-    fi
+install: composer update --no-progress --prefer-dist
 
 matrix:
+  include:
+    - name: PHP 7.0 Unit Tests
+      php: 7.0
+      install:
+        - composer require php-coveralls/php-coveralls:^2.2 --dev --no-update
+        - composer update --no-progress --prefer-dist
+      script: vendor/bin/phpunit --coverage-clover build/logs/clover.xml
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
Travis seems to have a bug where it reads a stale config file. Closing and opening to re-trigger.

GrahamCampbell:
// cc @TomasVotruba

GrahamCampbell:
Ready for review (and merge) now @nikic. :)

nikic:
@GrahamCampbell I'm waiting for the PHPUnit release on Friday.

GrahamCampbell:
Ah, ok. I will remove the dev stability lines I added, on Friday, then. :)

GrahamCampbell:
Rebased just now. :)

GrahamCampbell:
image

.travis.yml
```diff
@@ -1,35 +1,43 @@
 language: php
 dist: xenial
 
-cache:
-  directories:
-    - $HOME/.composer/cache
+before_install: composer self-update --2
```
nikic, line 4: Is this a requirement for anything?

.travis.yml
```diff
@@ -1,35 +1,43 @@
...
-
-install:
-  - if [ $TRAVIS_PHP_VERSION = '7.0' ]; then composer require satooshi/php-coveralls '~1.0'; fi
-  - |
-    if [ $TRAVIS_PHP_VERSION = 'nightly' ]; then
-      composer install --prefer-dist --ignore-platform-reqs;
-    else
-      composer install --prefer-dist;
-    fi
+install: composer update --no-progress --prefer-dist
```
nikic, line 6: Why update and not install?

.travis.yml
```diff
@@ -1,35 +1,43 @@
...
+    - name: PHP 7.2 Unit Tests
+      php: 7.2
+    - name: PHP 7.3 Unit Tests
+      php: 7.3
+    - name: PHP 7.4 Unit Tests
+      php: 7.4
+    - name: PHP 8.0 Unit Tests
+      php: nightly
+      install:
+        - composer require phpunit/phpunit:^9.3 --dev --no-update
```
nikic, line 28: This shouldn't be needed.

test_old/run-php-src.sh
```diff
@@ -1,5 +1,9 @@
-VERSION="master"
+if [[ $1 == '7' ]]; then
+	VERSION='PHP-7.4'
+else
+	VERSION='master'
+fi
```
nikic, line 5: It would be better to fetch the tarball for specific tags here.

test_old/run.php
```diff
@@ -14,7 +14,7 @@ function showHelp($error) {
 
     php run.php [--no-progress] testType pathToTestFiles
 
-The test type must be one of: PHP5, PHP7 or Symfony.
+The test type must be one of: PHP5, PHP7, PHP8 or Symfony.
```
nikic, line 17: I'd suggest using test type `PHP` only and adding a `--php-version` option, that allows specifying the exact version to use (and pass on to the lexer).

.travis.yml
```diff
@@ -1,35 +1,43 @@
...
+      php: 7.4
+      script:
+        - test_old/run-php-src.sh 7
+    - name: PHP 8 Integration Tests
+      php: 7.4
+      script:
+        - test_old/run-php-src.sh 8
   allow_failures:
-    - php: nightly
+    - name: PHP 8.0 Unit Tests
```
nikic, line 39: Now that PHPUnit is compatible, there's no need to allow failures for PHP 8 unit tests anymore.

.travis.yml
GrahamCampbell: Because install is only allowed when there is a lock file. It has always been the case, but composer 1 doesn't complain loudly. Composer 2 does, and Composer 3 will fail all together.

.travis.yml
GrahamCampbell: Allows us to remove the caching from travis, since composer 2 downloads in parallel. The end result is faster builds and less CO2 emissions.

GrahamCampbell: It also allows us to only ignore the php platform requirement, instead of all platform requirements, on PHP 8.

NB Ignoring that is still needed because not all of PHPUnit's dependencies allow PHP 8 yet, even though phpunit does.

test_old/run-php-src.sh
GrahamCampbell: Yeh, maybe. I was going to do the latest tags for both versions, but they are too old, actually. We'd need PHP 7.4.10 and 8.0.0 beta 2 or 3, when they are tagged,

test_old/run.php
GrahamCampbell: I have just implemented the minimum changes needed to allow us to test what we want. If you'd like to improve the script with what you described, that'd be fine, of course, however I won't get the time to do that.

test_old/run-php-src.sh
GrahamCampbell: I've pinned to the current PHP-7.4 and master states, for now.

.travis.yml
```diff
@@ -26,16 +26,18 @@ matrix:
...
-      php: 7.4
+    - name: PHP 7.3 Code on PHP 8.0 Integration Tests
+      php: nightly
+      install:
+        - composer update --ignore-platform-req=php --no-progress --prefer-dist
       script:
-        - test_old/run-php-src.sh 7
-    - name: PHP 8 Integration Tests
-      php: 7.4
+        - test_old/run-php-src.sh 7.3.21
```
GrahamCampbell, line 34: That version of php doesn't have the php-src patch in the zend tests yet. No tagged php 7.3 does, as I said in the last review comment. ;)

nikic:
Merged as https://github.com/nikic/PHP-Parser/commit/544aee1671ce978126571e5d53cc04c04c610a24 and https://github.com/nikic/PHP-Parser/commit/722119502f9a4f7ad1d6080e2d2a89981f5c6a8d. Thanks for working on this!

GrahamCampbell:
👍

PR #703, opened 2020-08-23
FPPP failing tests

ondrejmirtes:
A regression after fixing https://github.com/nikic/PHP-Parser/pull/699. The empty line in the try block is removed.


nikic:
Thanks for the test case, this should be fixed by https://github.com/nikic/PHP-Parser/commit/4bc82432dec0780315a7900500b79931be7153f2.

ondrejmirtes:
@nikic Perfect, thank you!

PR #705, opened 2020-08-25
FPPP failing test case with class nodes example

weaverryan:
Hi!

I found a small regression after #699, so sha: bd722809f73f84a8025ba368ae2fd056f562a6b7

In this case, the line break is lost between the original `private $id` and `public function getId()`. To trigger this, you appear to need to replace a node (index 2, `getFoo()) *and* add another new node after that. This triggers the lost line break on the earlier nodes.

Thanks!

### DIFF
--- a/test/code/formatPreservation/listRemoval.test
+++ b/test/code/formatPreservation/listRemoval.test
@@ -174,3 +174,36 @@ try {
 } catch (\Throwable $e) {
 
 }
+-----
+<?php
+class Foo
+{
+    private $id;
+
+    public function getId()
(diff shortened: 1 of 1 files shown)


nikic:
Thank you, this should be fixed by https://github.com/nikic/PHP-Parser/commit/ad365b1beb4d12d78c0eed489b070a83c43b854f. Maybe it's actually right this time around...

weaverryan:
Thank you - things are looking clean again where we use the library - you are the best! ❤️

PR #719, opened 2020-09-30
PrettyPrinter\\Standard: fixes nested ternaries output;

### LINKED ISSUE #718: PrettyPrinter brokes nested ternaries
Code like:

```php
<?php ( $a ? $b : $c ) ? $d : $e;
```

Is parsed correctly:

```
array(
    0: Stmt_Expression(
        expr: Expr_Ternary(
            cond: Expr_Ternary(
                cond: Expr_Variable( name: a )
                if: Expr_Variable( name: b )
                else: Expr_Variable( name: c )
            )
            if: Expr_Variable( name: d )
            else: Expr_Variable( name: e )
        )
    )
)
```

But `PrettyPrinter` will outputs:

```php
<?php $a ? $b : $c ? $d : $e;
```

So if you run the code it will gives:

```
Deprecated: Unparenthesized `a ? b : c ? d : e` is deprecated. Use either `(a ? b : c) ? d : e` or `a ? b : (c ? d : e)`
```

rentalhost:
Fixes #718 

That fixes it, but I don't know if it's the best way.

I also had to modify one of the tests, because it is "positively affected" by this commit. The previous expected output would generate the error mentioned in #718.

### DIFF
--- a/lib/PhpParser/PrettyPrinter/Standard.php
+++ b/lib/PhpParser/PrettyPrinter/Standard.php
@@ -659,6 +659,12 @@ class Standard extends PrettyPrinterAbstract
     }
 
     protected function pExpr_Ternary(Expr\Ternary $node) {
+        if ($node->cond instanceof Expr\Ternary) {
+            return '(' . $this->pInfixOp(Expr\Ternary::class,
+                $node->cond, ') ?' . (null !== $node->if ? ' ' . $this->p($node->if) . ' ' : '') . ': ', $node->else
+            );
+        }
+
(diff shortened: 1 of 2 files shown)


nikic:
Thanks for the PR! I've applied a simpler fix in https://github.com/nikic/PHP-Parser/commit/eff72eeffa1dee6e04ca6c3d49e3a46e30e638dd. This is just a matter of marking the operator as non-associative.

rentalhost:
Oh! I almost got it right on my first attempts! I have tried `1` but it brokes other tests. I didn't notice that `0` could be tested as well. Haha!

Thanks a lot for this quick response!

PR #720, opened 2020-10-01
Assign could be an AssignOp

rentalhost:


nikic:
While `Assign` and `AssignOp` have nominally the same structure, their semantics are different: `Assign` is used for `=`, while `AssignOp` is used for `op=` ("compound assign"). `instanceof AssignOp` is the way to determine whether something is a compound assignment, so having `Assign` implement that would break the semantics.

PR #729, opened 2020-11-02
Introduce node scalar escaped to avoid wrong escaping - Close #725

### LINKED ISSUE #725: String with `Y-m-d\TH:i:sP` is outputted with double backslashes
I use the latest PHP-Parser version `v4.10.2` and I want to generate a class constant with the value/string of `DATE_ATOM` (`Y-m-d\TH:i:sP`). This is a simple test script:

```php
final class ClassConstant extends PhpParser\NodeVisitorAbstract
{
    public function enterNode(PhpParser\Node $node)
    {
        if ($node instanceof PhpParser\Node\Stmt\Class_) {
            $node->stmts[] = new PhpParser\Node\Stmt\ClassConst(
                [
                    new PhpParser\Node\Const_(
                        'OUTPUT_FORMAT',
                        new PhpParser\Node\Scalar\String_('Y-m-d\TH:i:sP') // DATE_ATOM, same behavior with double quotes
                    ),
                ]
            );

            return $node;
        }

        return null;
    }
}

$parser = (new PhpParser\ParserFactory())->create(PhpParser\ParserFactory::ONLY_PHP7);
$printer = new PhpParser\PrettyPrinter\Standard(['shortArraySyntax' => true]);

$ast = $parser->parse('<?php final class DateTimeVO {}');

$nodeTraverser = new PhpParser\NodeTraverser();
$nodeTraverser->addVisitor(new ClassConstant());

print_r($printer->prettyPrintFile($nodeTraverser->traverse($ast)));
``` 

I would expect an output like
```php
<?php

final class DateTimeVO
{
    const OUTPUT_FORMAT = 'Y-m-d\TH:i:sP';
}
```
but I get instead
```php
<?php

final class DateTimeVO
{
    const OUTPUT_FORMAT = 'Y-m-d\\TH:i:sP'; // double \\
}
```
Is this a bug or is there another way to solve it? Something like to force the output of the value as is?

sandrokeil:
See #725 for more details. 

It's needed for code generation where the value should not be escaped.

### DIFF
--- a/lib/PhpParser/Node/Scalar/Escaped_.php
+++ b/lib/PhpParser/Node/Scalar/Escaped_.php
@@ -1,0 +1,30 @@
+<?php declare(strict_types=1);
+
+namespace PhpParser\Node\Scalar;
+
+use PhpParser\Node\Scalar;
+
+class Escaped_ extends Scalar
+{
+    /** @var string String value */
+    public $value;
+
+    /**
+     * Constructs a string scalar node.
+     *
+     * @param string $value      Value of the string
+     * @param array  $attributes Additional attributes
+     */
+    public function __construct(string $value, array $attributes = []) {
+        $this->attributes = $attributes;
+        $this->value = $value;
+    }
+
+    public function getSubNodeNames() : array {
+        return ['value'];
+    }
+
+    public function getType() : string {
+        return 'Scalar_Escaped';
+    }
+}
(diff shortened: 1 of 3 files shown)


sandrokeil:
Solved by https://github.com/nikic/PHP-Parser/commit/cf0cd6003e797f7d10eb7e8fe7937341d88d6c93

PR #732, opened 2020-11-07
Fixed type vs string

quisse:
Php parses `{` as an array `[0 => 386,1 => '{',2 => 12,]` so this check doesn't work properly. My edit fixes this.

### DIFF
--- a/lib/PhpParser/Internal/TokenStream.php
+++ b/lib/PhpParser/Internal/TokenStream.php
@@ -46,7 +46,8 @@ class TokenStream
      * @return bool
      */
     public function haveBraces(int $startPos, int $endPos) : bool {
-        return $this->haveTokenImmediatelyBefore($startPos, '{')
+        return ($this->haveTokenImmediatelyBefore($startPos, '{')
+                || $this->haveTokenImmediatelyBefore($startPos, T_CURLY_OPEN))
             && $this->haveTokenImmediatelyAfter($endPos, '}');
     }
 
@@ -201,6 +202,7 @@ class TokenStream
 
     public function haveBracesInRange(int $startPos, int $endPos) {
         return $this->haveTokenInRange($startPos, $endPos, '{')
+            || $this->haveTokenInRange($startPos, $endPos, T_CURLY_OPEN)
             || $this->haveTokenInRange($startPos, $endPos, '}');
     }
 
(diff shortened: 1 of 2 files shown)


quisse:
Proof can be found here https://3v4l.org/WRjDC

lib/PhpParser/Internal/TokenStream.php
```diff
@@ -64,6 +64,7 @@ public function haveTokenImmediatelyBefore(int $pos, $expectedTokenType) : bool
         $tokens = $this->tokens;
         $pos--;
         for (; $pos >= 0; $pos--) {
+            $index = is_string($expectedTokenType)?1:0;
```
nikic, line 67: The $index variable is not used, so this cannot possibly have an impact on the behavior. Please, can you add a test case for this issue in https://github.com/nikic/PHP-Parser/tree/master/test/code/formatPreservation?

quisse: How do I run those test cases?

quisse:
Tests are succeeding. Could you have another look please?

lib/PhpParser/Internal/TokenStream.php
```diff
@@ -64,12 +64,19 @@ public function haveTokenImmediatelyBefore(int $pos, $expectedTokenType) : bool
...
+            } else {
+                $tokenType = $token;
+            }
             if ($tokenType === $expectedTokenType) {
                 return true;
             }
-            if ($tokenType !== \T_WHITESPACE
-                && $tokenType !== \T_COMMENT && $tokenType !== \T_DOC_COMMENT) {
+            if (
+                $tokenType !== \T_WHITESPACE && $tokenType !== \T_COMMENT && $tokenType !== \T_DOC_COMMENT && $tokenType !== ' '
```
nikic, line 78: This doesn't look right to me, as this is now specific to one particular kind of whitespace.

Taking a step back here, I think it might be better not to modify the haveTokenImmediatelyBefore/After helpers. Instead, the calls can be adjusted.

I assume in this case the problem is haveBraces(). We can replace the call `$this->haveTokenImmediatelyBefore($startPos, '{')` with `$this->haveTokenImmediatelyBefore($startPos, '{') || $this->haveTokenImmediatelyBefore($startPos, T_CURLY_OPEN)`.

quisse: Yes. the issue indeed is in the `haveBraces` method. I'll fix it there. `haveBracesInRange` will also have this issue I suppose.

PR #745, opened 2020-12-22
Add .github directory to .gitattributes export-ignore

Ayesh:
Adds `.github` directory to `.gitattributes` file with an `export-ignore` rule.
This also aligns all `export-ignore` rules for more readability.

Thank you.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,9 +1,9 @@
-/doc export-ignore
-/test export-ignore
-/test_old export-ignore
-.gitattributes export-ignore
-.gitignore export-ignore
-.travis.yml export-ignore
-CHANGELOG.md export-ignore
+/.github         export-ignore
+/doc             export-ignore
+/test            export-ignore
+/test_old        export-ignore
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -1,9 +1,10 @@
...
-.gitignore export-ignore
-.travis.yml export-ignore
-CHANGELOG.md export-ignore
+/.github         export-ignore
+/doc             export-ignore
+/test            export-ignore
+/test_old        export-ignore
+.gitattributes   export-ignore
+.gitignore       export-ignore
+.travis.yml      export-ignore
```
nikic, line 7: The .travis.yml line can be dropped now.

Ayesh: You are right, thank you. I force-pushed with new changes. 

 - Adds `.github` directory to `.gitattributes` file with an `export-ignore` rule.
 - This also aligns all `export-ignore` rules for more readability.
 - Removes `.travis.yml` file that is no longer used.

Thank you.

PR #748, opened 2020-12-31
Fixed branch alias

GrahamCampbell:


composer.json
```diff
@@ -22,7 +22,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "4.9-dev"
+            "dev-master": "4.10-dev"
```
nikic, line 25: Is it possible to make this `4.x-dev`?

GrahamCampbell: Pinning the minor version is better so that composer knows the difference between `~4.9.0@dev`, `~4.10.0@dev` , `~4.11.0@dev`.

GrahamCampbell: Alternatively, naming the branch `4.10` avoids needing to use and update the aliases at all.

nikic: I'm happy to remove the branch-alias altogether if 4.x-dev is not acceptable.

PR #750, opened 2021-01-08
Fix bug in pretty printer

danog:
Starting from PHP 8, a number of additional expressions and non-expressions can appear on the rhs of an instanceof expression, including a string (appropriately wrapped in parenthesis).

### DIFF
--- a/lib/PhpParser/PrettyPrinter/Standard.php
+++ b/lib/PhpParser/PrettyPrinter/Standard.php
@@ -638,7 +638,7 @@ class Standard extends PrettyPrinterAbstract
             . '(' . $this->pCommaSeparated($node->params) . ')'
             . (null !== $node->returnType ? ': ' . $this->p($node->returnType) : '')
             . ' => '
-            . $this->p($node->expr);
+            . $this->pDereferenceLhs($node->expr);
     }
 
     protected function pExpr_ClosureUse(Expr\ClosureUse $node) {
--- a/lib/PhpParser/PrettyPrinter/Standard.php
+++ b/lib/PhpParser/PrettyPrinter/Standard.php
@@ -1032,8 +1032,11 @@ class Standard extends PrettyPrinterAbstract
     }
 
     protected function pNewVariable(Node $node) {
-        // TODO: This is not fully accurate.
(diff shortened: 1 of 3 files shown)


lib/PhpParser/PrettyPrinter/Standard.php
```diff
@@ -1036,6 +1036,14 @@ protected function pNewVariable(Node $node) {
         return $this->pDereferenceLhs($node);
     }
 
+    protected function pInstanceOfClass(Node $node) {
+        if (!$this->callLhsRequiresParens($node) && !$node instanceof Scalar\String_) {
+            return $this->p($node);
+        } else  {
+            return '(' . $this->p($node) . ')';
+        }
+    }
```
nikic, line 1045: This should modify pNewVariable(), as the rules for new and instanceof are the same.

danog: Well, in this case, they aren't the same, as doing `new ("a")();` still isn't allowed in PHP 8.

danog:
BTW, also added a similar fix for arrow functions, to avoid problems when printing things like fn () => ($a + $b);

lib/PhpParser/PrettyPrinter/Standard.php
```diff
@@ -638,7 +638,7 @@ protected function pExpr_ArrowFunction(Expr\ArrowFunction $node) {
             . '(' . $this->pCommaSeparated($node->params) . ')'
             . (null !== $node->returnType ? ': ' . $this->p($node->returnType) : '')
             . ' => '
-            . $this->p($node->expr);
+            . $this->pDereferenceLhs($node->expr);
```
nikic, line 641: I don't understand why this is needed. Can you share an example for arrow functions where parentheses are needed, but currently not added? The test changes are incorrect (in that the added parentheses are redundant).

nikic:
This is fixed by https://github.com/nikic/PHP-Parser/commit/1eb6b5653eb7aeb9e12d2bbbaa8b8f698528e0a2 and https://github.com/nikic/PHP-Parser/commit/cc34c2450c50504b7b3aaa36dfd6276eb9be77d8.

Issue #757, opened 2021-02-06
[PHP 8.1] Add enumarations


TomasVotruba:
- RFC: https://wiki.php.net/rfc/enumerations
- PHP core PR: https://github.com/php/php-src/pull/6489/files

TomasVotruba:
Working on it :)

PR #758, opened 2021-02-06
[PHP 8.1] Add Enums

TomasVotruba:
Closes https://github.com/nikic/PHP-Parser/issues/757


szepeviktor:
`[WIP]` is really a Draft PR on GitHub :)

kép

TomasVotruba:
I know, thanks :) I've seen more people forgot to switch that than to make it useful.

test/code/parser/stmt/class/enum.test
```diff
@@ -0,0 +1,20 @@
...
+        )
+        name: Identifier(
+            name: Foo
+        )
+        implements: array(
+        )
+        stmts: array(
+        )
+    )
+)
```
TomasVotruba, line 20: This test now fails on unexpected `T_STRING`:

```bash
-'array(
-    0: Stmt_Enum(
-        attrGroups: array(
+'Syntax error, unexpected T_STRING from 3:6 to 3:8
+array(
+    0: Stmt_Expression(
+        expr: Expr_ConstFetch(
+            name: Name(
+                parts: array(
+                    0: enum
+                )
+            )
         )
-        name: Identifier(
-            name: Foo
-        )
-        implements: array(
-        )
-        stmts: array(
+    )
+    1: Stmt_Expression(
+        expr: Expr_ConstFetch(
+            name: Name(
+                parts: array(
+                    0: Foo
+                )
+            )
         )
     )
 )'
```

@nikic Any hint where is the problematic grammar?

How to run only this case?

```bash
vendor/bin/phpunit test/PhpParser/CodeParsingTest.php --filter "*enum*"
```

TomasVotruba:
@nikic I made first draft, but got stuck on unclear error probably related to `enum` name grammar. I'm lost about why `enum` is not taken as keyword.

Could you review the PR so far? Thank you

szepeviktor:
> Hi, I made first ~~draft~~, but got stuck on

Hi, I made first `[WIP]`, but got stuck on ... 😃

TomasVotruba:
@nikic Hi, since Enums are now accepted and merged in core, could you give me feedback on this?
Thank you :+1:

nikic:
Can you please rebase over https://github.com/nikic/PHP-Parser/commit/a8223f228a7560954f01a0501504b623f1b47531? The enum keyword needs slightly special handling.

TomasVotruba:
Thanks! I've rebased and grammar works perfectly :+1: 

I'm adding more tests case and there is weird one, where `string` is not interpreted as `Identifier`, but as `Name`. Any how to fix it?

The failing tests report:

image


Otherwise it's ready to review

nikic:
@TomasVotruba Possibly this is because you added enums to the PHP 5 grammar as well, where `string` is a class type? I would suggest to drop them from php5.y and mark tests as `!!php7`.

TomasVotruba:
I've removed enum from php5.y grammer and it looks good :+1:

lib/PhpParser/Node/Stmt/EnumCase.php
```diff
@@ -0,0 +1,37 @@
...
+     */
+    public function __construct($name, $expr, array $subNodes = [], array $attributes = []) {
+        parent::__construct($attributes);
+        $this->name = \is_string($name) ? new Node\Identifier($name) : $name;
+        $this->expr = $expr;
+        $this->attrGroups = $subNodes['attrGroups'] ?? [];
+    }
+
+    public function getSubNodeNames() : array {
+        return ['name', 'expr', 'attrGroups'];
```
nikic, line 31: attrGroups should come first in this array (order of code).

lib/PhpParser/Node/Stmt/EnumCase.php
```diff
@@ -0,0 +1,37 @@
...
+    public $expr;
+    /** @var Node\AttributeGroup[] PHP attribute groups */
+    public $attrGroups;
+
+    /**
+     * @param string|Node\Identifier    $name
+     * @param Node\Expr|null            $expr
+     * @param array                     $attributes Additional attributes
+     * @param array                     $subNodes   Array of the following optional subnodes:
+     *                                      'attrGroups' => array() : PHP attribute groups
```
nikic, line 21: As there's only one optional parameter here, I think we should accept `array $attrGroups = []` directly, instead of going through subNodes.

lib/PhpParser/Node/Stmt/Enum_.php
```diff
@@ -0,0 +1,40 @@
...
+
+        parent::__construct($attributes);
+    }
+
+    public function getSubNodeNames() : array {
+        return ['attrGroups', 'name', 'scalarType', 'implements', 'stmts'];
+    }
+
+    public function getType() : string {
+        return 'Stmt_Class';
```
nikic, line 38:
```suggestion
        return 'Stmt_Enum';
```

lib/PhpParser/ParserAbstract.php
```diff
@@ -940,7 +941,7 @@ protected function checkClass(Class_ $node, $namePos) {
     protected function checkInterface(Interface_ $node, $namePos) {
         if (null !== $node->name && $node->name->isSpecialClassName()) {
             $this->emitError(new Error(
-                sprintf('Cannot use \'%s\' as class name as it is reserved', $node->name),
+                sprintf('Cannot use \'%s\' as interface name as it is reserved', $node->name),
```
nikic, line 944: https://3v4l.org/P6c7v The message uses `class name`.

lib/PhpParser/PrettyPrinter/Standard.php
```diff
@@ -737,6 +737,14 @@ protected function pStmt_Trait(Stmt\Trait_ $node) {
              . $this->nl . '{' . $this->pStmts($node->stmts) . $this->nl . '}';
     }
 
+    protected function pStmt_EnumCase(Stmt\EnumCase $node) {
+        return $this->pAttrGroups($node->attrGroups)
+            . 'case ' . $node->name
+            . ($node->expr ? ' = ' . $this->p($node->expr) : '')
+            . ';'
+        ;
+    }
```
nikic, line 746: Also need a pStmt_Enum, right?

test/code/formatPreservation/enum.test
```diff
@@ -0,0 +1,23 @@
...
+!!php7
+<?php
+
+enum Size {
+    case Small;
+        case Medium     =     1000;
+
+            case Large;
+            case ExtraLarge;
+}
```
nikic, line 23: Missing a test for the normal pretty-printer.

lib/PhpParser/Node/Stmt/EnumCase.php
TomasVotruba: Makes sense

lib/PhpParser/ParserAbstract.php
TomasVotruba: Ah, I see

lib/PhpParser/PrettyPrinter/Standard.php
TomasVotruba: Yes :+1:

test/code/formatPreservation/enum.test
```diff
@@ -14,10 +14,10 @@ $stmts[0]->stmts[] = new Node\Stmt\EnumCase('ExtraLarge', null);
...
+{
     case Small;
-        case Medium     =     1000;
-
-            case Large;
-            case ExtraLarge;
+    case Medium     =     1000;
+    case Large;
+    case ExtraLarge;
 }
```
TomasVotruba: Not sure why, but after adding `pStmt_Enum()` the spacing broke here

TomasVotruba:
Ready to review :heavy_check_mark:

TomasVotruba:
I've processed your comments :+1: 
Let me know if there is something missing.

TomasVotruba:
Thanks :partying_face:

PR #764, opened 2021-03-10
Fixed corruption of hex escaped characters in the pretty printer

### LINKED ISSUE #763: PHP-Parser does not properly handle UTF-8 multibyte characters such as U+00C0
Hi,
I've encountered an issue where php-parser corrupts hex escaped characters in PHP code. I've made a repro in a [gist](https://gist.github.com/marekdedic/9ef501e2dcc898b9fd0272e25d2e0116) - run `run.php` and diff source and output to see the issue. Somehow deleting anything from this file changes the output of those unicode characters, so it may be file-size related?

Let me know if you need any additional info.

PHP version: `7.4.15`
php-parser version: `v4.10.4`

marekdedic:
Closes #763


nikic:
I'm not sure I understand the issue you're fixing. PHP strings are binary, they do not need to be valid UTF-8.

It would make sense to me to print non-UTF-8 portions of strings hex-encoded, but this isn't what you're doing here. Unless I'm missing something, this is also going to convert all UTF-8 continuation bytes to hex, which is probably not desired.

nikic:
Added a test in https://github.com/nikic/PHP-Parser/commit/38aa0920c97e8e85ad4c9d0a6b03236a7db17a52 to make sure that has coverage.

marekdedic:
Hi,
honestly, I'm just trying to get #763 fixed (see the included gist for better description of the issue). If you know of a better way to do that, I would appreciate any pointers :)

marekdedic:
See https://github.com/nikic/PHP-Parser/issues/763#issuecomment-805022196

PR #765, opened 2021-03-21
Adding class constant to builder factory

whizsid:


TomasVotruba:
I actually needed this couple of times :+1:

lib/PhpParser/Builder/ClassConst.php
```diff
@@ -0,0 +1,103 @@
...
+{
+    protected $flags = 0;
+    protected $attributes = [];
+    protected $constants = [];
+
+    /**
+     * Creates a class constant builder
+     *
+     * @param string|Identifier                          $name  - Name
+     * @param Node\Expr|bool|null|int|float|string|array $value — Value
```
nikic, line 23:
```suggestion
     * @param string|Identifier                          $name  Name
     * @param Node\Expr|bool|null|int|float|string|array $value Value
```
These dashes shouldn't be there (same below).

lib/PhpParser/Builder/ClassConst.php
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * Add another constant to const group
+     *
+     * @param string|Identifier                          $name  - Name
+     * @param Node\Expr|bool|null|int|float|string|array $value — Value
+     *
+     * @return $this The builder instance (for fluid interface)
+     */
+    public function addConst($name, $value) {
+        array_push($this->constants, new Const_($name, BuilderHelpers::normalizeValue($value)));
```
nikic, line 38:
```suggestion
        $this->constants[] = new Const_($name, BuilderHelpers::normalizeValue($value));
```

lib/PhpParser/Builder/ClassConst.php
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * Returns the built class node.
+     *
+     * @return Stmt\Property The built constant node
+     */
+    public function getNode(): PhpParser\Node {
+        return new Stmt\ClassConst(
+            $this->constants,
+            $this->flags !== 0 ? $this->flags : Stmt\Class_::MODIFIER_PUBLIC,
```
nikic, line 99: I believe this should just pass `$this->flags`, to use implicitly public constants if `makePublic()` is not called. Otherwise there is no way to create implicitly public constants at all.

lib/PhpParser/Builder/ClassConst.php
```diff
@@ -0,0 +1,103 @@
...
+            'comments' => [BuilderHelpers::normalizeDocComment($docComment)]
+        ];
+
+        return $this;
+    }
+
+    /**
+     * Returns the built class node.
+     *
+     * @return Stmt\Property The built constant node
```
nikic, line 94:
```suggestion
     * @return Stmt\ClassConst The built constant node
```

lib/PhpParser/BuilderFactory.php
```diff
@@ -161,6 +161,18 @@ public function useConst($name) : Builder\Use_ {
         return new Builder\Use_($name, Use_::TYPE_CONSTANT);
     }
 
+    /**
+     * Creates a class constant builder.
+     *
+     * @param string|Identifier                          $name  - Name
+     * @param Node\Expr|bool|null|int|float|string|array $value — Value
+     *
+     * @return Builder\Use_ The created use const builder
```
nikic, line 170:
```suggestion
     * @return Builder\ClassConst The created use const builder
```

test/PhpParser/Builder/ClassConstTest.php
```diff
@@ -0,0 +1,166 @@
...
+    public function testModifiers() {
+        $node = $this->createClassConstBuilder("TEST", 1)
+            ->makePrivate()
+            ->getNode()
+        ;
+
+        $this->assertEquals(
+            new Stmt\ClassConst(
+                [
+                    new Const_("TEST", new LNumber(1) )
```
nikic, line 28:
```suggestion
                    new Const_("TEST", new LNumber(1))
```
and below

test/PhpParser/BuilderFactoryTest.php
```diff
@@ -36,6 +36,11 @@ public function provideTestFactory() {
         ];
     }
 
+    public function testFactoryClassConst(){
```
nikic, line 39:
```suggestion
    public function testFactoryClassConst() {
```

PR #781, opened 2021-06-10
Add Attribute builder and support for named args

simivar:
In this PR I've added two builders for named arguments and attributes which later could be  used for Class/Properties builders

### DIFF
--- a/lib/PhpParser/BuilderFactory.php
+++ b/lib/PhpParser/BuilderFactory.php
@@ -23,7 +23,7 @@ class BuilderFactory
     public function attribute($name, array $args = []) : Node\Attribute {
         return new Node\Attribute(
             BuilderHelpers::normalizeName($name),
-            $this->namedArgs($args)
+            $this->args($args)
         );
     }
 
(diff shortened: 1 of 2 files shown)


lib/PhpParser/BuilderFactory.php
```diff
@@ -12,6 +12,21 @@
...
+     *
+     * @param string|Name $name Name of the attribute
+     * @param array            $args Attribute named arguments
+     *
+     * @return Node\Attribute
+     */
+    public function attribute($name, array $args = []) : Node\Attribute {
+        return new Node\Attribute(
+            BuilderHelpers::normalizeName($name),
+            $this->namedArgs($args)
```
nikic, line 26: Why do attributes required named args in particular?

simivar: You're right that it's not a requirement, I've seen most of the examples around the web that use named arguments for readability. How would you resolve this? Move to normal args? Add some other parameter to the method?

nikic: I think you should be able to support this in the normal args() method. If the key is a string, treat as named, otherwise as positional. That way it's supported everywhere, and optional.

simivar: Great idea, thank you. I've changed the code accordingly.

test/PhpParser/BuilderFactoryTest.php
```diff
@@ -96,6 +107,18 @@ public function testArgs() {
         );
     }
 
+    public function testNamedArgs() {
+        $factory = new BuilderFactory();
+        $unpack = new Arg(new Expr\Variable('bar'), false, true, [], new Identifier('foo'));
```
nikic, line 112: This doesn't really make sense in that `foo: ...$bar` is not legal. We don't necessarily have to guard against that, but at least shouldn't use that as a test case. In this case, I'd test something like `['foo', 'bar' => 'baz']` just to have the mix of positional and named.

lib/PhpParser/BuilderFactory.php
```diff
@@ -12,6 +12,21 @@
 
 class BuilderFactory
 {
+    /**
+     * Creates an attribute node.
+     *
+     * @param string|Name $name Name of the attribute
+     * @param array            $args Attribute named arguments
```
nikic, line 19:
```suggestion
     * @param array             $args Attribute named arguments
```

nikic (commented):
This looks good to me, just one nit on the test.

test/PhpParser/BuilderFactoryTest.php
simivar: Changed 👍

lib/PhpParser/BuilderFactory.php
simivar: Changed 👍

PR #782, opened 2021-06-11
Add addAttribute() method to Builders with normalizer

simivar:
Adds `addAttribute()` method to Builders of all nodes supporting attributes with `BuilderHelpers::normalizeAttribute()` usage inside so we can pass both `Node\Attribute` and `Node\AttributeGroup` instances

### DIFF
--- a/lib/PhpParser/BuilderHelpers.php
+++ b/lib/PhpParser/BuilderHelpers.php
@@ -284,7 +284,7 @@ final class BuilderHelpers
         }
 
         if (!($attribute instanceof Node\Attribute)) {
-            throw new \LogicException('Attribute must be an instance of PhpParser\Comment\Attribute or PhpParser\Comment\AttributeGroup');
+            throw new \LogicException('Attribute must be an instance of PhpParser\Node\Attribute or PhpParser\Node\AttributeGroup');
         }
 
         return new Node\AttributeGroup([$attribute]);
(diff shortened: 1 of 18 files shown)


nikic:
One thing I'm wondering here is whether it would make sense to make this `addAttribute()` instead which accepts either `Attribute` or `AttributeGroup` and normalizes the former to the latter. My thinking is that in most cases you want to add an attribute group with a single attribute, and we could save the boilerplate of constructing the AttributeGroup object in that case.

lib/PhpParser/BuilderHelpers.php
```diff
@@ -270,6 +270,26 @@ public static function normalizeDocComment($docComment) : Comment\Doc {
...
+     * @return Node\AttributeGroup The Attribute Group
+     */
+    public static function normalizeAttribute($attribute) : Node\AttributeGroup
+    {
+        if ($attribute instanceof Node\AttributeGroup) {
+            return $attribute;
+        }
+
+        if (!($attribute instanceof Node\Attribute)) {
+            throw new \LogicException('Attribute must be an instance of PhpParser\Comment\Attribute or PhpParser\Comment\AttributeGroup');
```
nikic, line 287:
```suggestion
            throw new \LogicException('Attribute must be an instance of PhpParser\Node\Attribute or PhpParser\Node\AttributeGroup');
```

simivar: Fixed, thank you 👍

simivar:
@nikic I think this is a great idea, was wondering how to deal with unnecessary repetition of `new AttributeGroup()`. All the changes are made now and PR can be reviewed 🎉

PR #788, opened 2021-06-18
Add support for union types in BuilderHelpers::normalizeType()

simivar:


nikic:
I'm not sure this is moving in the right direction. As PHP's type system is becoming more complex, I don't think we want to support creating complex types from string here. It's still kinda okay for union types, but if you go towards `A&B|C&D` you need to deal with things like precedence.

For union types, I think the type should be created using `$factory->unionType($type1, $type2, ...)` or similar.

Probably the nullable type case should have been handled this way as well, but I guess we can't just drop that now.

simivar:
I totally get your point and I've moved the implementation from `BuilderHelpers::normalizeType()` to `BuilderFactory::unionType()`. I've implemented it first in the `BuilderHelpers::normalizeType()` because it is used in, for example, `Builder\Param` so I wanted to support splitting to `UnionType` there.

lib/PhpParser/BuilderFactory.php
```diff
@@ -143,6 +143,49 @@ public function function(string $name) : Builder\Function_ {
...
+     *
+     * @return Node\UnionType The created union type
+     */
+    public function unionType(...$types) : Node\UnionType {
+        $numberOfTypes = count($types);
+        if ($numberOfTypes < 2) {
+            throw new \LogicException('Expected at least two types in the Union Type');
+        }
+
+        $standaloneTypes = [
```
nikic, line 159: Why is this reimplementing this logic rather than calling normalizeType() on all the $types?

lib/PhpParser/BuilderFactory.php
```diff
@@ -143,6 +143,49 @@ public function function(string $name) : Builder\Function_ {
...
+
+            if (in_array($type, $standaloneTypes, true)) {
+                throw new \LogicException(sprintf('%s can be only used as standalone type', $type));
+            }
+
+            if (strpos((string) $type, '?') !== false) {
+                throw new \LogicException('Union type should not contain nullable type, use null instead');
+            }
+
+            $normalizedType = BuilderHelpers::normalizeType($type);
```
nikic, line 178: I think this call should happen at the start of the loop. E.g. the above check for nullable types would make more sense as a check for NullableType, rather than a `?type` string specifically.

Though I'm not sure we should be including all these checks at all. PHP-Parser generally doesn't try hard to enforce compiler constraints, and there are no syntactic problems with creating `void|int` or whatever. I don't think we really need to go out of our way to check for this, especially as these checks aren't performed on the parser side (i.e. PHP-Parser will accept `void|int` without throwing an error).

nikic: Basically, I'd reduce this function to just:

```
return new Node\UnionType(array_map(['BuilderHelpers', 'normalizeType'], $types));
```

PR #789, opened 2021-06-18
Simplify BuilderHelpers::normalizeName() implementation

simivar:
In order to get rid of the flag in `BuilderHelpers::normalizeNameCommon()` I have moved all the logic related to the normalization of the name to the `BuilderHelpers::normalizeName()` method and expr-related stuff to the `BuilderHelpers::normalizeNameOrExpr()` method which later calls the basic `normalizeName()` as well

### DIFF
--- a/lib/PhpParser/BuilderHelpers.php
+++ b/lib/PhpParser/BuilderHelpers.php
@@ -104,29 +104,6 @@ final class BuilderHelpers
      * @return Name The normalized name
      */
     public static function normalizeName($name) : Name {
-        return self::normalizeNameCommon($name, false);
-    }
-
-    /**
-     * Normalizes a name: Converts string names to Name nodes, while also allowing expressions.
-     *
-     * @param Expr|Name|string $name The name to normalize
-     *
-     * @return Name|Expr The normalized name or expression
-     */
-    public static function normalizeNameOrExpr($name) {
-        return self::normalizeNameCommon($name, true);
-    }
-
-    /**
-     * Normalizes a name: Converts string names to Name nodes, optionally allowing expressions.
-     *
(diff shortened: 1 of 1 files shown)


lib/PhpParser/BuilderHelpers.php
```diff
@@ -147,16 +124,28 @@ private static function normalizeNameCommon($name, bool $allowExpr) {
...
+     * @param Expr|Name|string $name The name to normalize
+     *
+     * @return Name|Expr The normalized name or expression
+     */
+    public static function normalizeNameOrExpr($name) {
+        if ($name instanceof Expr) {
+            return $name;
+        }
+
+        if (!is_string($name) && !($name instanceof Name)) {
```
nikic, line 142: This creates some coupling, in that this check needs to stay in sync with the types supported by normalizeName().

I think a possibly better way to go about this would be to have a tryNormalizeName() method that doesn't throw an instead returns null on failure, and then have

```
public static function normalizeName($name) {
    $name = self::tryNormalizeName($name);
    if ($name !== null) {
        return $name;
    }
    throw new \LogicException('Name must be a string or an instance of Node\Name');
}
```
and similar for the other method.

simivar: Wouldn't that as well introduce coupling between `normalizeNameOrExpr` and `tryNormalizeName` but not strictly with `if` statements but with the error message passed to the `\LogicException`? We would need to list all of the nodes that are supported by `tryNormalizeName` and add "or Node\Expr" in the `normalizeNameOrExpr`.

nikic: Hm yeah, there isn't really any optimal solution here, so let's go with your version!

PR #790, opened 2021-06-19
Add BuilderFactory::nullableType()

simivar:
As @nikic mentioned in the #788 it would be nice if nullable types could be handled in the BuilderFactory as their own function instead of additional checks in `BuilderHelpers::normalizeType()`. This is the first step to removing that support: providing an alternative. 

Later, we could deprecate `NullableType|UnionType` param types for `BuilderHelpers::normalizeType()` and remove it altogether in the next major version (e.g. 5).

### DIFF
--- a/lib/PhpParser/BuilderFactory.php
+++ b/lib/PhpParser/BuilderFactory.php
@@ -360,6 +360,36 @@ class BuilderFactory
     }
 
     /**
+     * Creates a Nullable Type out of a non-normalized type
+     *
+     * @param string|Name|Identifier|Node\NullableType $type
+     *
+     * @return Node\NullableType
+     */
+    public function nullableType($type) : Node\NullableType {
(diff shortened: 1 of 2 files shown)


lib/PhpParser/BuilderFactory.php
```diff
@@ -359,6 +359,36 @@ public function concat(...$exprs) : Concat {
...
+
+        if (!is_string($type) && !($type instanceof Name) && !($type instanceof Identifier)) {
+            throw new \LogicException('Type must be a string, or an instance of Name, Identifier or NullableType');
+        }
+
+        if (strlen($type) > 0 && $type[0] === '?') {
+            $type = substr($type, 1);
+        }
+
+        $notNullableTypes = [
```
jurchiks, line 382: Make this static.

PR #792, opened 2021-07-03
Add handling for Enum(Case)s in NameResolver

bwoebi:


nikic:
Can you please also add them to the test at https://github.com/nikic/PHP-Parser/blob/c35cc4b2cb82fd543444d6afeb42f858b5e375e3/test/PhpParser/NodeVisitor/NameResolverTest.php#L180?

bwoebi:
Sure, done - and added the missing type printing to the pretty printer at the same time

lib/PhpParser/PrettyPrinter/Standard.php
```diff
@@ -731,6 +731,7 @@ protected function pStmt_Enum(Stmt\Enum_ $node) {
         return $this->pAttrGroups($node->attrGroups)
              . 'enum ' . $node->name
              . (!empty($node->implements) ? ' implements ' . $this->pCommaSeparated($node->implements) : '')
+			 . ($node->scalarType ? " : $node->scalarType" : '')
```
nikic, line 734: Tabs...

test/PhpParser/NodeVisitor/NameResolverTest.php
```diff
@@ -327,6 +339,7 @@ public function testAddDeclarationNamespacedName() {
             ]),
             new Stmt\Trait_('E'),
             new Expr\New_(new Stmt\Class_(null)),
+			new Stmt\Enum_('F'),
```
nikic, line 342: ...tabs

bwoebi:
Sorry for the tabs … I wished PHPStorm would just properly recognize the indentation style used in a file and not try to put tabs inmidst of spaces :-/

nikic:
Looks like test/code/prettyPrinter/stmt/enum.test needs an update now that the type is printed.

bwoebi:
Oh, there are also non-PHP files, fixed.

PR #796, opened 2021-07-15
Add support for new PHP 8.1 modifiers

kocsismate:


grammar/php7.y
```diff
@@ -535,6 +535,7 @@ optional_visibility_modifier:
     | T_PUBLIC                  { $$ = Stmt\Class_::MODIFIER_PUBLIC; }
     | T_PROTECTED               { $$ = Stmt\Class_::MODIFIER_PROTECTED; }
     | T_PRIVATE                 { $$ = Stmt\Class_::MODIFIER_PRIVATE; }
+    | T_READONLY                { $$ = Stmt\Class_::MODIFIER_READONLY }
```
kocsismate, line 538: gotcha! :D

staabm: Missing semicolon?
```suggestion
    | T_READONLY                { $$ = Stmt\Class_::MODIFIER_READONLY; }
```

kocsismate: Yes, that is the main problem, but there are a few other smaller issues as well which cause a some test failures.

PR #797, opened 2021-07-20
[PrettyPrinter] Failing test Pretty Printer should not duplicate comment

samsonasik:
Given the following code:

```php
    if ( true ) :
        // TEST 1
      elseif ( true ) :
        // TEST 2
        // TEST 3
        // TEST 4
      endif;
```

It currently print back:

```diff
    if (true) {
        // TEST 1
    } elseif (true) {
        // TEST 2
        // TEST 3
        // TEST 4
+       // TEST 2
+       // TEST 3
+       // TEST 4
    }
```

which duplicate on elseif part.

### DIFF
--- a/test/code/prettyPrinter/comments.test
+++ b/test/code/prettyPrinter/comments.test
@@ -65,3 +65,27 @@ function test()
 {
     // empty
 }
+-----
+<?php
+
+function noDuplicateComment()
+{
+    if ( true ) :
+        // TEST 1
+      elseif ( true ) :
+        // TEST 2
+        // TEST 3
+        // TEST 4
+      endif;
+}
+-----
+function noDuplicateComment()
+{
+    if (true) {
+        // TEST 1
+    } elseif (true) {
+        // TEST 2
+        // TEST 3
+        // TEST 4
+    }
+}


samsonasik:
The issue seems because of using `:` with `endif` syntax, when changed to `{ }`, it is working as expected.

nikic:
To clarify, this works fine with the normal pretty printer, but the comment gets duplicated if formatting preservation is enabled. The dump looks like this:
```
array(
    0: Stmt_Function[3:1 - 10:1](
        attrGroups: array(
        )
        byRef: false
        name: Identifier[3:10 - 3:27](
            name: noDuplicateComment
        )
        params: array(
        )
        returnType: null
        stmts: array(
            0: Stmt_If[5:5 - 9:10](
                cond: Expr_ConstFetch[5:9 - 5:12](
                    name: Name[5:9 - 5:12](
                        parts: array(
                            0: true
                        )
                    )
                )
                stmts: array(
                    0: Stmt_Nop[6:0 - 6:17](
                        comments: array(
                            0: // TEST 1
                        )
                    )
                )
                elseifs: array(
                    0: Stmt_ElseIf[7:5 - 7:18](
                        cond: Expr_ConstFetch[7:13 - 7:16](
                            name: Name[7:13 - 7:16](
                                parts: array(
                                    0: true
                                )
                            )
                        )
                        stmts: array(
                            0: Stmt_Nop[8:0 - 8:17](
                                comments: array(
                                    0: // TEST 2
                                )
                            )
                        )
                        comments: array(
                            0: // TEST 1
                        )
                    )
                )
                else: null
            )
        )
    )
)
```
Presumably, the issue is that the `Nop` with the comments is located outside the `Endif` node.

nikic:
Fixed by https://github.com/nikic/PHP-Parser/commit/9b46dffb12c6c83d33b993ae81b9a2895293dee8.

PR #798, opened 2021-07-21
Type hint that namespacedName may be null

mglaman:
The namedspacedName may be null if not set when NameResolver is not used


nikic:
Can it be null if the property exists?

mglaman:
@nikic I was running code which passed `namespacedName` to a function and it came back null – no other errors. I guess maybe the appropriate thing is to call `isset` versus checking if the value is null?

TomasVotruba:
Fixed and replace by already merged https://github.com/nikic/PHP-Parser/commit/f4b835f7d858d90b58c6d965a7d0f87d2ba57c43

This can be closed

PR #805, opened 2021-09-09
Print comma before comments for new array item (fixes #804)

### LINKED ISSUE #804: When printing array with added items, comments are added to the previous item
When I have an array like this:

```php
[
  /**
   * Some comment
   */
  'a' => 'value',
]
```

and I parse the following text, extract the `ArrayItem` node for `b` and append it to the previous AST:

```php
[
  /**
   * Some comment
   */
  'a' => 'value',

  /**
   * Something something
   */
  'b' => 'value',
]
```

And I print using format-preserving, I get:

```php
[
  /**
   * Some comment
   */
  'a' => 'value'
  /**
   * Something something
   */,
  'b' => 'value',
]
```

Note that the comma marking the end of `a` is placed after b's doc comment. Not only is this technically wrong:
1. It makes the operation irreversible. If you parse the new printed file, the doc comment will be lost, since it doesn't belong to `a` or `b`.
2. It leads to invalid PHP code, in the case of line comments, as the terminating comma for `a` is commented out:
   ```php
   'a' => 'value'
   // Something something,
   'b' => 'value',
   ```

From my observations, this happens:
1. With all types of comments (single-line, multiline, doc)
2. Whether or not there is a trailing comma after `a`
3. Whether or not there is a blank line between a and b
4. Regardless of any of the line or tokenPos attributes on the comments or items

When I inspect the nodes, everything is parsed correctly. The comment is identified as belonging to `b`. It's only during printing that the comment is printed out-of-order.

shalvah:
This fixes #804 by printing the `$insertStr` before the comments, in the case of items added to multiline arrays.

### DIFF
--- a/lib/PhpParser/PrettyPrinterAbstract.php
+++ b/lib/PhpParser/PrettyPrinterAbstract.php
@@ -824,7 +824,11 @@ abstract class PrettyPrinterAbstract
                     return null;
                 }
 
-                if ($insertStr === ', ' && $this->isMultiline($origNodes)) {
+                // We go multiline if the original code was multiline,
+                // or if it's an array item with a comment above it.
+                if ($insertStr === ', ' && (
+                        $this->isMultiline($origNodes) || !empty($arrItem->getAttribute('comments')))
+                ) {
                     $insertStr = ',';
                     $insertNewline = true;
                 }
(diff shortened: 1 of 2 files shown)


nikic:
Could you please add a test for this? https://github.com/nikic/PHP-Parser/blob/master/test/code/formatPreservation/listInsertion.test shows how to do that. Basically original code, followed by transform script followed by expected output.

shalvah:
Sure👍. Was looking for how to do that, but then got distracted.

shalvah:
Added tests 👍

While doing that, discovered that two-item arrays were still being rendered as single-line even with a comment on one of them, thanks to https://github.com/nikic/PHP-Parser/blob/0f0e52a95e5d9822f2ad5f728ac87c344636de95/lib/PhpParser/PrettyPrinterAbstract.php#L1093-L1095, so fixed that.

PR #808, opened 2021-09-15
Fixed doccomment in ArrowFunction

kukulich:


lib/PhpParser/Node/Expr/ArrowFunction.php
```diff
@@ -20,7 +20,7 @@ class ArrowFunction extends Expr implements FunctionLike
     /** @var null|Node\Identifier|Node\Name|Node\ComplexType */
     public $returnType;
 
-    /** @var Expr */
+    /** @var Expr|null */
```
nikic, line 23: This looks like a bug in the constructor ... it should require `expr` to be specified.

kukulich: There is `$this->expr = $subNodes['expr'] ?? null;` now...

nikic: Yes, and there shouldn't be...

kukulich: Fixed.

PR #816, opened 2021-11-05
Added builders for enum and enum case

kukulich:


lib/PhpParser/Builder/EnumCase.php
```diff
@@ -0,0 +1,85 @@
...
+{
+    protected $name;
+    protected $value = null;
+    protected $attributes = [];
+
+    /** @var Node\AttributeGroup[] */
+    protected $attributeGroups = [];
+
+    /**
+     * Creates a enum case builder
```
nikic, line 23:
```suggestion
     * Creates am enum case builder.
```

lib/PhpParser/Builder/EnumCase.php
```diff
@@ -0,0 +1,85 @@
...
+     * @param string|Identifier $name  Name
+     */
+    public function __construct($name) {
+        $this->name = $name;
+    }
+
+    /**
+     * Sets the value.
+     *
+     * @param Node\Expr $value
```
nikic, line 34:
```suggestion
     * @param Node\Expr|string|int $value
```

lib/PhpParser/BuilderFactory.php
```diff
@@ -71,6 +71,17 @@ public function trait(string $name) : Builder\Trait_ {
         return new Builder\Trait_($name);
     }
 
+    /**
+     * Creates a enum builder.
```
nikic, line 75:
```suggestion
     * Creates an enum builder.
```

lib/PhpParser/BuilderFactory.php
```diff
@@ -188,6 +199,17 @@ public function classConst($name, $value) : Builder\ClassConst {
         return new Builder\ClassConst($name, $value);
     }
 
+    /**
+     * Creates a enum case builder.
```
nikic, line 203:
```suggestion
     * Creates an enum case builder.
```

test/PhpParser/Builder/EnumCaseTest.php
```diff
@@ -0,0 +1,98 @@
...
+                new Scalar\String_('Hallo World')
+            ],
+            [
+                [1, 2, 3],
+                new Expr\Array_([
+                    new Expr\ArrayItem(new Scalar\LNumber(1)),
+                    new Expr\ArrayItem(new Scalar\LNumber(2)),
+                    new Expr\ArrayItem(new Scalar\LNumber(3)),
+                ])
+            ],
```
nikic, line 95: While we don't explicitly validate against them, float and array aren't actually allowed here, so I wouldn't test them.

PR #826, opened 2022-01-25
Change print order of modifiers

Bl00D4NGEL:
(First contribution so please bare with me :) )

While looking through the issues for rector I noticed that the following error (https://github.com/rectorphp/rector/issues/6963) is caused by the printer of this package. While it is perfectly valid to have `public abstract` in the code I would argue that it makes more sense to switch them up to be `abstract public`. 

This is achieved by changing the order in which the modifiers are printed. I figured that both final and abstract could/should be printed before the the public/protected/private modifier.


nikic:
I think this change makes sense for better PSR compatibility, but I don't think it can be merged into PHP-Parser 4.x due to the BC break (I expect it will break tests in dependent projects). I'd recommend that Rector override the pModifier() method with the preferred order for now (it's protected).

PR #828, opened 2022-02-24
fix PrettyPrinter AttributeGroup print error

LinkedDestiny:
Fix PrettyPrinter to make it work in PHP8.0

### DIFF
--- a/lib/PhpParser/PrettyPrinter/Standard.php
+++ b/lib/PhpParser/PrettyPrinter/Standard.php
@@ -67,7 +67,11 @@ class Standard extends PrettyPrinterAbstract
     }
 
     protected function pAttributeGroup(Node\AttributeGroup $node) {
-        return '#[' . $this->pCommaSeparated($node->attrs) . ']';
+        $result = [];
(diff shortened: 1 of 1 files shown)


nikic:
I'm not sure I understand what this is supposed to fix. This seems like a formatting change?

LinkedDestiny:
> I'm not sure I understand what this is supposed to fix. This seems like a formatting change?

when i use this in PHP8.0, the generated code is invalid like this:

`
Attribute()
public const ENUM_EXAMPLE = 1
`

The '#[]' is not present.

LinkedDestiny:
> I'm not sure I understand what this is supposed to fix. This seems like a formatting change?

image

Here is the main change.

nikic:
Can you please add a test to https://github.com/nikic/PHP-Parser/tree/master/test/code/prettyPrinter?

othercorey:
@LinkedDestiny Do you plan to finish this?

nikic:
Closing this due to inactivity. I don't understand the issue this fixes from the code change, and there is no test case the demonstrates the problem.

PR #831, opened 2022-05-09
[String_] Add rawValue attribute

TomasVotruba:
In Rector we have to hack around tokens to get the original value, because the `parse()` method flattens original information:

https://github.com/nikic/PHP-Parser/blob/a6e34665fd6a2bcefc924cb5fd73788767ae510e/lib/PhpParser/Node/Scalar/String_.php#L55-L72

We could use also attribute value, but it would require more grammar changes.

Closes https://github.com/nikic/PHP-Parser/pull/577

Also fixes https://github.com/nikic/PHP-Parser/issues/576

### DIFF
--- a/grammar/php5.y
+++ b/grammar/php5.y
@@ -691,7 +691,6 @@ scalar_dereference:
       array_expr '[' dim_offset ']'                         { $$ = Expr\ArrayDimFetch[$1, $3]; }
     | T_CONSTANT_ENCAPSED_STRING '[' dim_offset ']'
           { $attrs = attributes();
-            $attrs['kind'] = strKind($1);
             $$ = Expr\ArrayDimFetch[Scalar\String_::fromString($1, $attrs), $3]; }
     | constant '[' dim_offset ']'                           { $$ = Expr\ArrayDimFetch[$1, $3]; }
     | scalar_dereference '[' dim_offset ']'                 { $$ = Expr\ArrayDimFetch[$1, $3]; }
@@ -797,7 +796,6 @@ common_scalar:
     | T_DNUMBER                                             { $$ = Scalar\DNumber::fromString($1, attributes()); }
     | T_CONSTANT_ENCAPSED_STRING
           { $attrs = attributes();
-            $attrs['kind'] = strKind($1);
             $$ = Scalar\String_::fromString($1, $attrs, false); }
     | T_LINE                                                { $$ = Scalar\MagicConst\Line[]; }
     | T_FILE                                                { $$ = Scalar\MagicConst\File[]; }
--- a/grammar/php5.y
+++ b/grammar/php5.y
@@ -692,8 +692,7 @@ scalar_dereference:
     | T_CONSTANT_ENCAPSED_STRING '[' dim_offset ']'
           { $attrs = attributes();
             $attrs['kind'] = strKind($1);
-            $attrs['rawValue'] = $1;
-            $$ = Expr\ArrayDimFetch[new Scalar\String_(Scalar\String_::parse($1), $attrs), $3]; }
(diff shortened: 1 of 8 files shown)


TomasVotruba:
@todo add test

dryabov:
But `get(Start|End)FilePos` is a standard way to get access to raw data, e.g. well-known [`psalm`](https://github.com/vimeo/psalm) works in this way (see [`Psalm\Internal\Provider\StatementsProvider::parseStatements`](https://github.com/vimeo/psalm/blob/4.x/src/Psalm/Internal/Provider/StatementsProvider.php#L440-L453) for parser initialization there, and search its source code for `getStartFilePos()` and `getAttribute('startFilePos')` to get usage examples).

TomasVotruba:
The change of value is the issue.

TomasVotruba:
Added `rawValue` attribute :heavy_check_mark:

grammar/php5.y
```diff
@@ -690,7 +690,9 @@ array_expr:
 scalar_dereference:
       array_expr '[' dim_offset ']'                         { $$ = Expr\ArrayDimFetch[$1, $3]; }
     | T_CONSTANT_ENCAPSED_STRING '[' dim_offset ']'
-          { $attrs = attributes(); $attrs['kind'] = strKind($1);
+          { $attrs = attributes();
+            $attrs['kind'] = strKind($1);
+            $attrs['rawValue'] = $1;
```
nikic, line 695: Should this also use a fromString() approach?

TomasVotruba: I'd preffer that, but did not want to change the API much :+1: 

Will handle..

TomasVotruba:
Updated and rebased :+1: Ready for review :heavy_check_mark:

grammar/php5.y
```diff
@@ -690,8 +690,9 @@ array_expr:
 scalar_dereference:
       array_expr '[' dim_offset ']'                         { $$ = Expr\ArrayDimFetch[$1, $3]; }
     | T_CONSTANT_ENCAPSED_STRING '[' dim_offset ']'
-          { $attrs = attributes(); $attrs['kind'] = strKind($1);
-            $$ = Expr\ArrayDimFetch[new Scalar\String_(Scalar\String_::parse($1), $attrs), $3]; }
+          { $attrs = attributes();
+            $attrs['kind'] = strKind($1);
```
nikic, line 694: We'd want to move the strKind() logic into fromString() as well. Similar to how LNumber kind is also determined in LNumber::fromString().

TomasVotruba: Done :heavy_check_mark:

grammar/php5.y
```diff
@@ -691,7 +691,6 @@ scalar_dereference:
       array_expr '[' dim_offset ']'                         { $$ = Expr\ArrayDimFetch[$1, $3]; }
     | T_CONSTANT_ENCAPSED_STRING '[' dim_offset ']'
           { $attrs = attributes();
```
nikic, line 693: Can directly pass attributes() now, without the extra var.

TomasVotruba: Done :heavy_check_mark:

TomasVotruba:
Thanks :+1:

nikic:
Probably this should be added to EncapsedStringPart as well?

TomasVotruba:
What exactly do you mean? Have `fromString()` method there?

nikic:
I mean having a rawValue attribute. It seems inconsistent to only add it to non-interpolated strings.

TomasVotruba:
I'm on it :wink:

TomasVotruba:
See https://github.com/nikic/PHP-Parser/pull/837

PR #832, opened 2022-05-09
[LNumber] Add rawValue attribute to LNumber to allow numeric separator etc.

TomasVotruba:
Allows to use `$number->rawValue` to work with numberic seaprator https://github.com/nikic/PHP-Parser/pull/615

Now we have to work with actual tokens on the position to get the original value.
When accepted, I will add the same feature for `DNumber`.

Replaces https://github.com/nikic/PHP-Parser/pull/655

### DIFF
--- a/lib/PhpParser/Node/Scalar/LNumber.php
+++ b/lib/PhpParser/Node/Scalar/LNumber.php
@@ -16,20 +16,15 @@ class LNumber extends Scalar
     /** @var int Number value */
     public $value;
 
-    /** @var int|string The string type is used for numeric separator, e.g. 1_000 */
-    public $rawValue;
-
     /**
      * Constructs an integer number scalar node.
      *
-     * @param int             $value      Value of the number
-     * @param array           $attributes Additional attributes
-     * @param string|int|null $rawValue
+     * @param int   $value      Value of the number
+     * @param array $attributes Additional attributes
      */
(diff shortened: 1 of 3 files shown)


TomasVotruba:
With this node, we could avoid token + original token hacking to change to node value:
https://github.com/rectorphp/rector-src/pull/2302/files#diff-635a41f527b2bd76a9c3701b11fdd4c28979b500ccb26f57baa062f45bd316d1

TomasVotruba:
Ready for review :+1:

nikic:
This should be added as an attribute, not as a subnode.

TomasVotruba:
I can update it 👍 

How can we reprint the value (e.g. without "_") without removing the orig node attribute?

TomasVotruba:
> This should be added as an attribute, not as a subnode.

Done :heavy_check_mark:

nikic:
> How can we reprint the value (e.g. without "_") without removing the orig node attribute?

Removing the orig node attribute is the correct way to reprint the value. Why do you want to avoid that?

TomasVotruba:
>  Why do you want to avoid that?

It always seems little hacky when I do that in Rector.  I think "If it does not change the node, remove original node". 
But I don't fully understand the logic behind it.

TomasVotruba:
Thanks :+1:

PR #833, opened 2022-05-09
[DNumber] Add rawValue attribute to hold the original value

TomasVotruba:
Mirror to https://github.com/nikic/PHP-Parser/pull/832

### DIFF
--- a/lib/PhpParser/Node/Scalar/DNumber.php
+++ b/lib/PhpParser/Node/Scalar/DNumber.php
@@ -9,15 +9,20 @@ class DNumber extends Scalar
     /** @var float Number value */
     public $value;
 
+    /** @var float|string */
+    public $rawValue;
+
     /**
      * Constructs a float number scalar node.
      *
-     * @param float $value      Value of the number
-     * @param array $attributes Additional attributes
+     * @param float             $value      Value of the number
+     * @param array             $attributes Additional attributes
+     * @param string|float|null $rawValue
      */
-    public function __construct(float $value, array $attributes = []) {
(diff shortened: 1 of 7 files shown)


TomasVotruba:
This PR will help using native node properties, instead of token + attribute hacking:
https://github.com/rectorphp/rector-src/pull/2302

lib/PhpParser/Node/Scalar/DNumber.php
```diff
@@ -24,6 +24,17 @@ public function getSubNodeNames() : array {
...
+     * @param mixed[] $attributes
+     */
+    public static function fromString(string $str, array $attributes = []): DNumber
+    {
+        $attributes['rawValue'] = $str;
+        $float = self::parse($str);
+
+        return new DNumber($float, $attributes);
+    }
+
```
TomasVotruba, line 37: This method should mirror the `LNumber::fromString()` https://github.com/nikic/PHP-Parser/pull/832/files#diff-dcef01e803010221926e1c1f1d388bd86cd219dc843253abf177cca48fb57ba1

We could add this to grammar itself, but it would require either custom method in `AbstractParser` or some grammar coding. That would be both harder to maintain and inconsistent with `LNumber`.

TomasVotruba:
The value is now in `"rawValue"` attribute.

Ready for review :heavy_check_mark:

TomasVotruba:
:clap:

TomasVotruba:
With this feature, tokens are now finally removed from  @rectorphp 🥳 
https://github.com/rectorphp/rector-src/pull/2321/files

PR #834, opened 2022-05-12
[PHP 8.2] Add readonly class support

TomasVotruba:
Ref https://wiki.php.net/rfc/readonly_classes

PHP src: https://github.com/php/php-src/pull/7305


TomasVotruba:
Ready for review :+1:

test/code/parser/stmt/class/modifier.test
```diff
@@ -66,7 +66,7 @@ array(
     )
 )
 -----
-<?php class A { readonly readonly $a; }
+<?php class C { readonly readonly $a; }
```
nikic, line 69: What's up with these changes?

grammar/php7.y
```diff
@@ -384,6 +384,7 @@ class_entry_type:
       T_CLASS                                               { $$ = 0; }
     | T_ABSTRACT T_CLASS                                    { $$ = Stmt\Class_::MODIFIER_ABSTRACT; }
     | T_FINAL T_CLASS                                       { $$ = Stmt\Class_::MODIFIER_FINAL; }
+    | T_READONLY T_CLASS                                    { $$ = Stmt\Class_::MODIFIER_READONLY; }
```
nikic, line 387: This looks like a bug in the php-src implementation to me: https://github.com/php/php-src/pull/7305/files#r871755250

nikic (commented):
This is missing a pretty printer test (and maybe pretty printer implementation -- not sure if this works out of the box).

test/code/parser/stmt/class/modifier.test
TomasVotruba: The diff for this modifier test was showing invalid mask on class with name A.
There are 3 classes with name A, so it took a while to figure out, which case it is and how to fix it. Unique name should make this easier.

grammar/php7.y
TomasVotruba: I though so too, but the RFC does not mention word about `abstract` / `final`, so I assumed it's unique and mutually exclusive.

Let's see.

TomasVotruba:
> This is missing a pretty printer test (and maybe pretty printer implementation -- not sure if this works out of the box).

I've added simple print. Seems working as the modifier method is universal.

grammar/php7.y
nikic: Actually, I misread the php-src implementation. It does allow `final readonly`, so we should also allow it here.

TomasVotruba: I could not find it in the tests there. Where did you find it?

I'll add it here. So `final readonly`? What about `abstract readonly`?

TomasVotruba: There it is, little hidden: https://github.com/php/php-src/pull/7305/files#diff-8bbc8315717cefd854b454c385beb1673710bbe9d0c7c07a89c6ebcacff1efbb

grammar/php7.y
```diff
@@ -382,8 +382,18 @@ enum_case_expr:
...
+class_modifiers:
+      class_modifier                                        { $$ = $1; }
+    | class_modifiers class_modifier                        { $this->checkModifier($1, $2, #2); $$ = $1 | $2; }
+;
+
+class_modifier:
+      T_ABSTRACT                                            { $$ = Stmt\Class_::MODIFIER_ABSTRACT; }
+    | T_FINAL                                               { $$ = Stmt\Class_::MODIFIER_FINAL; }
+    | T_READONLY                                            { $$ = Stmt\Class_::MODIFIER_READONLY; }
 ;
```
TomasVotruba, line 398: It's my rough guess how the grammer for 1-n modifiers should look like.
The `checkModifier()` method handles the invalid combinations.

TomasVotruba:
Ready for review :heavy_check_mark:

grammar/php7.y
nikic: We should probably add a separate method for class modifiers to produce the same error messages as https://github.com/php/php-src/blob/9a90bd705483004c2ef408ee9e9bb0902beade3f/Zend/zend_compile.c#L788. Currently "Cannot use the final modifier on an abstract class member" is used, which doesn't make sense when talking about a class rather than class member.

TomasVotruba:
I've added the custom check for class modifiers.

Let me know if that's ok. Otherwise good to review :heavy_check_mark:

lib/PhpParser/Node/Stmt/Class_.php
```diff
@@ -81,6 +81,27 @@ public function isAnonymous() : bool {
...
+        }
+
+        if ($a & self::MODIFIER_READONLY && $b & self::MODIFIER_READONLY) {
+            throw new Error('Multiple readonly modifiers are not allowed');
+        }
+
+        if ($a & 48 && $b & 48) {
+            throw new Error('Cannot use the final modifier on an abstract class');
+        }
+    }
```
nikic, line 103: Now that I see it, maybe just add an extra `bool $isClass` parameter to verifyModifier() to adjust the one error message that differs?

TomasVotruba: I was thinking on it, but then the method naming and purpose gets little weird. 
The `Class_::verifyModifier()` does not check class, but property/method/constant by default. 

Then there is also used in modifier adding in here: https://github.com/nikic/PHP-Parser/blob/3bf0082455cd6457453f6caaf25072684131448f/lib/PhpParser/BuilderHelpers.php#L310

That has no idea about it's parent - if class or property/method/constant. So the error messages get incompatible.



Current split of 2 methods is more compatible with the link you've sent me in php-src. Personally I'd go that way, as that's the easier way to add changes as php-src change.

nikic: For the BuilderHelper method, I think we do need the information about whether it's a class or not there as well, I think that currently these calls may produce an incorrect error message: https://github.com/nikic/PHP-Parser/blob/49e9951f2c1f504ecdaad0ab74df4b964c182afe/lib/PhpParser/Builder/Class_.php#L64-L84

TomasVotruba: I've added `[BuilderHelpers::addClassModifier()](https://github.com/nikic/PHP-Parser/pull/834/commits/e1526ce432be31993056a3c8a85e6d38391de878) to handle it :+1:

nikic:
Can you pleas regenerate the Php7.php file to fix the merge conflict?

TomasVotruba:
> Can you please regenerate the Php7.php file to fix the merge conflict?

Done :+1:

kocsismate:
@TomasVotruba Thanks for the implementation! I've just noticed a slight difference between how properties of a readonly class is handled by php-src and PHP-Parser: while the former automatically assigns the readonly flag to any such property, 
 PHP-Parser doesn't do so. The behavior of php-src is highlighted at https://github.com/php/php-src/pull/7305#pullrequestreview-714317994

Can we allow this inconsistency or should this be changed on PHP-Parser's part?

kocsismate:
> Saying that, I think php-parser is different from php-src, as the former parsers what it sees (and nothing implicit) and later covers implicit behavior not to allow invalid run. Like wysiwyg vs. runtime.

Aha, it sounds like a good rule of thumb, and makes sense to me.

> Maybe different point of views: What real php-parser use case is negatively affected by this?

Hmm, I'm not aware of any. Although, I'll have to adapt the readonly class support in my PR :)

P.S. @nikic Can you please make a PHP-Parser release so that I can update it in php-src?

PR #836, opened 2022-05-15
[Node] add StmtsIterable interface to mark nodes that contain iterable stmts to improve hooking in node visitors

TomasVotruba:
At the moment, there are nodes in php-parser, that include list of other stmts. While iterating over can be easy with 
`property_exists($node, 'stmts')` (feels like poor coding), the hooking to them via node visitors/Rector/PHPStan not so.

At the moment hooking into them requries special constant and careful listing of all such nodes:
https://github.com/rectorphp/rector-src/blob/975fdf113fab99b6120383211e997da2c820bd0a/rules/CodeQuality/NodeTypeGroup.php
It happened we forgot some and then Rector rules were not applied.

These rules often work with previously used stmt, e.g. to remove already assigned value:

```diff
 function setDefault()
 {
-    $value = 100;
     $value = 150;
     // ...
 }
```

This could be somehow achieved with next/previous/parent attributes. Yet PHPStan and Rector are moving away from next/previous/parent nodes for 2 reasons:

* it's heavy for memory
* it can take time to find firs previous stmts, as e.g. `Assign` and other `Expr` can be deeply nested:

```php
if ($number === 1000 && ($value = 100)++ > --$breakpoint) {
    
}
```

Here removing `$value = 100` itself is ilegal and requires quite complex check to all parent nodes. 
**Iterating only over all directly available stmts would make this no-brainer**.

---

### Where would be this interface be useful?

* Rector could hook into all those nodes via single type
* PHPStan that can hook only into single node (see `getNodeType()`) would be able to hook into those too 
* not just these 2 projects, but any php-parser based traverser could detect such node type easily

This marker will dramatically simplify integration of new rules to work with stmts.

<br>

At first I thought we could have `class StmtsAware extends Stmt {}`, but sometime `Expr` contains `$stmts` too:

https://github.com/nikic/PHP-Parser/blob/678ccbe0720549c51d30b76d094a117eac819e9e/lib/PhpParser/Node/Expr/Closure.php#L21-L22

Thus interface marker choice. We're already using such marker interface in Rector via patching of php-parser nodes: https://github.com/rectorphp/rector/blob/d08b83c4264ba5b265890a049285518f79684644/vendor/nikic/php-parser/lib/PhpParser/Node/Stmt/Foreach_.php#L8

And it works great :+1: 

<br>

What do you think?

### DIFF
--- a/lib/PhpParser/Internal/PrintableNewAnonClassNode.php
+++ b/lib/PhpParser/Internal/PrintableNewAnonClassNode.php
@@ -15,12 +15,7 @@ use PhpParser\Node\Expr;
  *
  * @internal
  */
-<<<<<<< HEAD
 class PrintableNewAnonClassNode extends Expr {
-=======
-class PrintableNewAnonClassNode extends Expr implements Node\StmtsIterable
-{
->>>>>>> 920aae4f (add StmtsIterable interface to mark nodes that contain iterable stmts to improve hooking in node visitors)
     /** @var Node\AttributeGroup[] PHP attribute groups */
     public $attrGroups;
     /** @var Node\Arg[] Arguments */
--- a/lib/PhpParser/Internal/PrintableNewAnonClassNode.php
+++ b/lib/PhpParser/Internal/PrintableNewAnonClassNode.php
@@ -15,7 +15,12 @@ use PhpParser\Node\Expr;
  *
  * @internal
  */
+<<<<<<< HEAD
 class PrintableNewAnonClassNode extends Expr {
+=======
+class PrintableNewAnonClassNode extends Expr implements Node\StmtsIterable
+{
+>>>>>>> 920aae4f (add StmtsIterable interface to mark nodes that contain iterable stmts to improve hooking in node visitors)
     /** @var Node\AttributeGroup[] PHP attribute groups */
     public $attrGroups;
     /** @var Node\Arg[] Arguments */
--- a/lib/PhpParser/Node/Stmt/ClassMethod.php
+++ b/lib/PhpParser/Node/Stmt/ClassMethod.php
@@ -6,7 +6,12 @@ use PhpParser\Modifiers;
 use PhpParser\Node;
 use PhpParser\Node\FunctionLike;
 
+<<<<<<< HEAD
 class ClassMethod extends Node\Stmt implements FunctionLike {
+=======
+class ClassMethod extends Node\Stmt implements FunctionLike, Node\StmtsIterable
+{
+>>>>>>> 920aae4f (add StmtsIterable interface to mark nodes that contain iterable stmts to improve hooking in node visitors)
     /** @var int Flags */
     public $flags;
     /** @var bool Whether to return by reference */
@@ -17,7 +22,7 @@ class ClassMethod extends Node\Stmt implements FunctionLike {
     public $params;
     /** @var null|Node\Identifier|Node\Name|Node\ComplexType Return type */
     public $returnType;
-    /** @var Node\Stmt[]|null Statements */
+    /** @var Node\Stmt[] Statements */
     public $stmts;
     /** @var Node\AttributeGroup[] PHP attribute groups */
     public $attrGroups;
--- a/lib/PhpParser/Node/Stmt/ClassMethod.php
+++ b/lib/PhpParser/Node/Stmt/ClassMethod.php
@@ -6,12 +6,7 @@ use PhpParser\Modifiers;
 use PhpParser\Node;
 use PhpParser\Node\FunctionLike;
 
-<<<<<<< HEAD
 class ClassMethod extends Node\Stmt implements FunctionLike {
-=======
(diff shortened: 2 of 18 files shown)


lib/PhpParser/Node/Stmt/ClassMethod.php
```diff
@@ -17,7 +17,7 @@ class ClassMethod extends Node\Stmt implements FunctionLike
     public $params;
     /** @var null|Node\Identifier|Node\Name|Node\ComplexType Return type */
     public $returnType;
-    /** @var Node\Stmt[]|null Statements */
+    /** @var Node\Stmt[] Statements */
     public $stmts;
```
TomasVotruba, line 21: This node uses `Stmt[]` from constructor as well:

```php
       //  'stmts'      => array()        : Statements
        $this->stmts = array_key_exists('stmts', $subNodes) ? $subNodes['stmts'] : [];
```

nikic: It can be null for abstract methods.

TomasVotruba: I see. Having this property nullable would disallow `StmtsIterable` consistency.

Can we update the grammar to pass  `[]` for the consistency?
https://github.com/nikic/PHP-Parser/blob/678ccbe0720549c51d30b76d094a117eac819e9e/grammar/php5.y#L519

The `abstract` method should be clear from `isAbstrat()` call:
https://github.com/nikic/PHP-Parser/blob/678ccbe0720549c51d30b76d094a117eac819e9e/lib/PhpParser/Node/Stmt/ClassMethod.php#L125-L127

Similar check for empty `ClassMethod` of the interface should be resolved from parent `Interface_` node.

What would you suggest?

lib/PhpParser/Node/StmtsIterable.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpParser\Node;
+
+use PhpParser\Node;
+
+/**
+ * @property Stmt[] $stmts
```
staabm, line 10: What about adding a getStmts() method instead and handle the `null` to `[]` case in the getter?

TomasVotruba: That could be also a way. How would you modify such stmts on such interface?

What do you think @nikic ?

TomasVotruba:
I feel like this got stuck but I don't know why. I'd love to get it before version 5 gets released.
What needs to be done here to finish it?

nikic:
> I feel like this got stuck but I don't know why. I'd love to get it before version 5 gets released. What needs to be done here to finish it?

This is blocked on the nullable `$stmts` in `ClassMethod`. I'm not sure what to do about it :/

lib/PhpParser/Internal/PrintableNewAnonClassNode.php
```diff
@@ -15,7 +15,7 @@
  *
  * @internal
  */
-class PrintableNewAnonClassNode extends Expr
+class PrintableNewAnonClassNode extends Expr implements Node\StmtsIterable
```
nikic, line 18: This one shouldn't implement it, for the same reason `Class_` etc don't.

lib/PhpParser/Node/Stmt/ClassMethod.php
TomasVotruba: I've removed the interface on the `ClassMethod`, as it does not comply with the conditions.

lib/PhpParser/Internal/PrintableNewAnonClassNode.php
TomasVotruba: Removed :+1:

TomasVotruba:
I've removed the interface from `ClassMethod`, as the use case is really different.   I also resolved the rest of comment.

The main goal of this interface is to **catch all the nested structures with single node**, and it does well now :+1: 

```php
if (...) {
    foreach (...) {
        $closure = function() {
            $item = 1;

            if (...) {
                return true;
            }
        };
    }
```

TomasVotruba:
Just curious, any chance to get this merged, except the `ClassMethod`?

Past 4 years we have to do lot of black magic autoload hacking, but it starts to falling apart while using with PHPUnit 12 and PHPStan tests (which is more and more cases). Ref https://github.com/rectorphp/rector-src/pull/7440

samsonasik:
`Node\Stmt\Block_` node can be `Node\StmtsIterable` as well

https://github.com/nikic/PHP-Parser/blob/0da2d6679a3df45d6d720aa2e0d4568f82a32e46/lib/PhpParser/Node/Stmt/Block.php#L7

this is direct `{}` without any holder, see https://3v4l.org/Kj358

TomasVotruba:
Continues in https://github.com/nikic/PHP-Parser/pull/1113/ without the `ClassMethod`

PR #844, opened 2022-07-04
Final public method for abstract class

MathiasReker:
All public methods of abstract classes should be final. Enforce API encapsulation in an inheritance architecture. If you want to override a method, use the Template method pattern.


nikic:
I don't agree with this guideline. I don't think "abstract" really factors into this -- if appropriate interfaces are provided one can make everything final, but I personally don't consider this to be worthwhile. And for these changes we usually don't provide interfaces either.

PR #849, opened 2022-07-04
Get class to class keyword

MathiasReker:
Replace get_class calls on object variables with class keyword syntax.


nikic:
This is only supported since PHP 8.0, while this library supports PHP >= 7.1.

PR #850, opened 2022-07-04
Use arrow functions

MathiasReker:
Anonymous functions with one-liner return statement must use arrow functions.


nikic:
This library supports PHP >= 7.1, while arrow functions are only available in PHP >= 7.4.

PR #857, opened 2022-07-04
Php unit dedicate assert

MathiasReker:
PHPUnit assertions like assertInternalType, assertFileExists, should be used over assertTrue.


nikic:
I don't trust PHPUnit's magic for this test (in particular, it is critically important that `instanceof` does not auto-load the class, and I have no idea whether PHPUnit does that or not).

Issue #471, opened 2018-02-05
Consider a different class for destructured array?


muglug:
Would it be feasible to create a separate class for array destructuring? Like `AssignmentArray_` or `ListArray_`.

Using a single class to represent two totally different scenarios (even though the syntax is very similar) is a bit of a challenge for static analysis.

TomasVotruba:
Do you have a use case?

muglug:
Oh, just the inconsistency that `Array_` on the LHS of an assignment may have items that are `null` (like `List_`), but an `Array_` on the RHS of an assignment is guaranteed to never have null items.

TomasVotruba:
Could you share PHP code? It's easier to spot and test

nikic:
I think the suggestion makes sense in general, but I'm going to punt on this for 4.0. I don't want to change types at this point.

muglug:
Makes sense – I'm about 12 months late with this

TomasVotruba:
Is this still relevant?

ondrejmirtes:
Yes

TomasVotruba:
How should the nodes look like for?

```php
[$a, $b,,, $c] = [1, 2, 3, 4, 5];
```

nikic:
What do you think about using `Expr\List_` for both `list($x, $y) = $z` and `[$x, $y] = $z`, together with an attribute that determines the actual style used (like `kind` on `Expr\Array_`)? That seems convenient in that you generally want to treat `list() =` and `[] =` the same, and it avoids an additional node type.

ondrejmirtes:
That would be really nice!

PR #859, opened 2022-07-20
PrettyPrinter: added option `shortListSyntax`

WinterSilence:
To print `[...]` instead of `list(...)`


nikic:
As this is a separate node, you can already control this by passing in either `Array` or `List` as the node type. I don't see a need for a pretty-printer option controlling this.

WinterSilence:
@nikic list is not array, like as attribute not comment(`#`). I think that such replacements are not the correct solution.

nikic:
This is the current representation though:
```
bin/php-parse '<?php [$x] = $y;'
====> Code <?php [$x] = $y;
==> Node dump:
array(
    0: Stmt_Expression(
        expr: Expr_Assign(
            var: Expr_Array(
                items: array(
                    0: Expr_ArrayItem(
                        key: null
                        value: Expr_Variable(
                            name: x
                        )
                        byRef: false
                        unpack: false
                    )
                )
            )
            expr: Expr_Variable(
                name: y
            )
        )
    )
)
```
`[$x]` will get represented as `Expr_Array` independently of whether it appears on the LHS or RHS. So at least as things are now, replacing the `Expr_List` node with an `Expr_Array` node would be the correct thing to do.

WinterSilence:
@nikic I'm understand you, but this is still not correct and may cause problems at printing

WinterSilence:
@nikic for example if `shortArraySyntax` is false, then `<?php [$x] = $y;` convert to invalid code `<?php array($x) = $y;` at printing

nikic:
> @nikic for example if `shortArraySyntax` is false, then `<?php [$x] = $y;` convert to invalid code `<?php array($x) = $y;` at printing

That's a good point. It may make sense to change the representation then (see also https://github.com/nikic/PHP-Parser/issues/471). A possibility is to invert things and always parse the LHS of assignments into a `List` rather than `Array` node, together with a flag that indicates whether `list()` or `[]` syntax was used.

WinterSilence:
> A possibility is to invert things and always parse the LHS of assignments into a List rather than Array node, together with a flag that indicates whether list() or [] syntax was used.

Yes, can you fix it? I not explored code of parsers yet

nikic:
With https://github.com/nikic/PHP-Parser/commit/68fc1ba4cb7b0aef9decc4e9d8f3b61eecdd3883 `List` is always used to represent destructuring and rendering is decided by `kind` attribute. The default is determined by `phpVersion`. I think that should address this use-case?

WinterSilence:
@nikic "short array destructuring" in PHP documentation called "symmetric array destructuring"

WinterSilence:
@nikic please, review/merge sended PR's before adding changes - resolve diff not so easy

nikic:
As I have already explained, I have no interest in supporting a separate option for this. You can already use phpVersion to control the default behavior, and flags on individual nodes to control precise behavior. The `shortArraySyntax` option exists for historical reasons only.

WinterSilence:
@nikic i explain why it's not flexible in #890

PR #861, opened 2022-07-21
Fix PhpParser\PrettyPrinter\Standard::pStmt_Function()

WinterSilence:
Same to `PhpParser\PrettyPrinter\Standard::pStmt_ClassMethod()` method


nikic:
Would be nice to add a pretty printer test for this.

nikic:
I also wonder whether we should do this for Expr_Closure as well? Wouldn't expect to hit that in practice, but probably makes sense to keep it consistent.

WinterSilence:
@nikic arrow functions is not short-cut to closure:
> except that using variables from the parent scope is always automatic

WinterSilence:
@nikic that's covered by `pStmt_Function()` tests

WinterSilence:
@nikic 
> I also wonder whether we should do this for Expr_Closure as well?

yes, but it's idea for another PR

PR #862, opened 2022-07-22
Add support for DNF types

Girgias:
This has nested arrays but I'm not sure I understand why the IntersectionType node is not pushed to the types array.

I reused the same grammar as is currently in php-src, so maybe that the issue?


grammar/php.y
```diff
@@ -703,14 +703,24 @@ type_without_static:
     | T_CALLABLE                                            { $$ = Node\Identifier['callable']; }
 ;
 
+union_type_element:
+                type { $$ = $1; }
+        |        '(' intersection_type ')' { $$ = $2; }
```
nikic, line 708: intersection_type is just an array at this point. You'd want to add an intermediate rule that turns it into a Node\IntersectionType.

grammar/php.y
```diff
@@ -703,14 +703,24 @@ type_without_static:
     | T_CALLABLE                                            { $$ = Node\Identifier['callable']; }
 ;
 
+union_type_element:
+                type { $$ = $1; }
+        |        '(' intersection_type ')' { $$ = Node\IntersectionType[$2]; }
```
nikic, line 708: This will assign incorrect position information (that includes the parentheses). It's necessary to do something like:
```
intersection_type:
      intersection_type_list { $$ = Node\IntersectionType[$1]; }
;
```
where `intersection_type_list` is the old `intersection_type`.

nikic (commented):
This also needs a test for the pretty printer -- as written, I'd expect the required parentheses to be missing.

grammar/php.y
Girgias: I'm not sure I understand why I need this, but maybe it because I'm rewriting the logic for ``pImplode()`` in ``pUnionType()`` as I didn't see what difference it makes. But then again there might be a better way of pretty printing then what I did.

lib/PhpParser/Node/UnionType.php
```diff
@@ -10,7 +10,7 @@ class UnionType extends ComplexType
     /**
      * Constructs a union type.
      *
-     * @param (Identifier|Name)[] $types      Types
+     * @param (Identifier|Name|IntersectinType)[] $types      Types
```
nikic, line 13:
```suggestion
     * @param (Identifier|Name|IntersectionType)[] $types      Types
```

PR #866, opened 2022-08-07
Adds class PhpParser\PrettyPrinter\Stub

WinterSilence:
Simple stub generator


WinterSilence:
@nikic ping

nikic:
I don't think this functionality should be part of this library. It's a specific application, which can be implemented on top. (Though I wouldn't implement it as a pretty-printer extension, it would make more sense to modify the AST to only leave the stub information behind.)

PR #867, opened 2022-08-07
Normalize comment classes

WinterSilence:
- `getReformattedText()`: fix `@return` types, remove comment trimming to keep empty lines in comment at begin and end, optimize detection of multiline comments
- `getShortestWhitespacePrefixLen()`: replace float `INF` to 0
- `jsonSerialize()`: update `@psalm-return` and remove redundant `@return`, fix `nodeType`

### DIFF
--- a/lib/PhpParser/Comment.php
+++ b/lib/PhpParser/Comment.php
@@ -147,16 +147,14 @@ class Comment implements \JsonSerializable
      * without trailing whitespace on the first line, but with trailing whitespace
      * on all subsequent lines.
      *
-     * @return mixed|string
+     * @return string|null
      */
     public function getReformattedText() {
-        $text = trim($this->text);
-        $newlinePos = strpos($text, "\n");
-        if (false === $newlinePos) {
+        if (0 < ($this->endLine - $this->startLine)) {
             // Single line comments don't need further processing
-            return $text;
+            return $this->text;
         }
-        if (preg_match('((*BSR_ANYCRLF)(*ANYCRLF)^.*(?:\R\s+\*.*)+$)', $text)) {
+        if (preg_match('((*BSR_ANYCRLF)(*ANYCRLF)^.*(?:\R\s+\*.*)+$)', $this->text)) {
             // Multi line comment of the type
             //
             //     /*
@@ -167,7 +165,7 @@ class Comment implements \JsonSerializable
             // is handled by replacing the whitespace sequences before the * by a single space
             return preg_replace('(^\s+\*)m', ' *', $this->text);
         }
-        if (preg_match('(^/\*\*?\s*[\r\n])', $text) && preg_match('(\n(\s*)\*/$)', $text, $matches)) {
+        if (preg_match('(^/\*\*?\s*[\r\n])', $this->text) && preg_match('(\n(\s*)\*/$)', $this->text, $matches)) {
             // Multi line comment of the type
             //
             //    /*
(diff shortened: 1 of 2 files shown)


nikic:
Test failures are legitimate.

WinterSilence:
@nikic yes, because
> remove comment trimming to keep empty lines in comment at begin and end

If you agree with these changes, I'll update the tests

nikic:
The change to trimming is fine -- to verify, I've landed it separately as https://github.com/nikic/PHP-Parser/commit/1f504d2c7da6020d56144cdb8443a3ba681f6183. This was a leftover from an older implementation, when comments were expected to have a trailing newline.

It's some of the other changes that cause problematic changes in formatting.

WinterSilence:
@nikic hm... I'm don't see how other changes can fail tests - content changed only here

lib/PhpParser/Comment.php
```diff
@@ -147,16 +147,14 @@ public function __toString() : string {
      * without trailing whitespace on the first line, but with trailing whitespace
      * on all subsequent lines.
      *
-     * @return mixed|string
+     * @return string|null
```
nikic, line 150: This can be just string. We can also make it a proper return type.

lib/PhpParser/Comment.php
```diff
@@ -209,7 +207,7 @@ public function getReformattedText() {
      */
     private function getShortestWhitespacePrefixLen(string $str) : int {
         $lines = explode("\n", $str);
-        $shortestPrefixLen = \INF;
+        $shortestPrefixLen = 0;
```
nikic, line 210: If this is set to 0, then the function always returns 0. If you want to avoid mixing of int / float, then I would suggest replacing with PHP_INT_MAX.

lib/PhpParser/Comment.php
```diff
@@ -221,14 +219,13 @@ private function getShortestWhitespacePrefixLen(string $str) : int {
...
 
     /**
-     * @return       array
-     * @psalm-return array{nodeType:string, text:mixed, line:mixed, filePos:mixed}
+     * @psalm-return array{nodeType: string, text: string, line: int, filePos: int, tokenPos: int, endLine: int,
+     * endFilePos: int, endTokenPos: int}
      */
     public function jsonSerialize() : array {
         // Technically not a node, but we make it look like one anyway
-        $type = $this instanceof Comment\Doc ? 'Comment_Doc' : 'Comment';
```
nikic, line 229: I'm not sure moving the logic into the other class really makes the code clearer...

lib/PhpParser/Comment.php
nikic: This probably causes the test failures.

lib/PhpParser/Comment.php
WinterSilence: you mean use `(string)` to returning values?

nikic: These preg_replace calls can only return string, so I think just adding `): string` to the method would be fine? I'm not sure how this ended up as `mixed|string`.

If this makes static analyzers complain because preg_replace can in theory also return array|null, then adding `(string)` would be fine to suppress.

lib/PhpParser/Comment.php
WinterSilence: fixed

WinterSilence:
@nikic tests failed because `startLine`/`endLine` not correct in current test

WinterSilence:
@nikic I'm fixed fake/incorrect data passing to `CommentTest` tests, I'm think other failed tests must be fixes in other PR's.

WinterSilence:
@nikic added alternative check to incomplete instances:
~~~php
|| ($this->endLine === -1 && strpos($this->text, "\n") !== false)
~~~

PR #869, opened 2022-08-08
[docs] Add generic types to NodeFinder to allow returning exact type in static analysis

TomasVotruba:
At the moment, the node finder does always returns generic `Node` or `Node[]` result:

```php
$nodeFinder = new NodeFinder();

// what is it? only Node[]
$returns = $nodeFinder->typeAwareNodeFinder->findInstanceOf($node, Return_::class);
```

In fact, we know it must be instances of `Return_`. To teach IDE/static analyzer this know-how, we have to add it manually in every result:

```php
/** @var Return_[] $resturn */
$returns = $nodeFinder->typeAwareNodeFinder->findInstanceOf($node, Return_::class);
```

This PR suggest to do the same already on `php-parser` level:

Screenshot from 2022-08-08 15-51-06

### DIFF
--- a/lib/PhpParser/NodeFinder.php
+++ b/lib/PhpParser/NodeFinder.php
@@ -31,11 +31,13 @@ class NodeFinder
 
     /**
      * Find all nodes that are instances of a certain class.
+
+     * @template TNode as Node
      *
-     * @param Node|Node[] $nodes Single node or array of nodes to search in
-     * @param string      $class Class name
+     * @param Node|Node[] $nodes    Single node or array of nodes to search in
+     * @param class-string<TNode>   $class Class name
      *
-     * @return Node[] Found nodes (all instances of $class)
+     * @return TNode[]              Found nodes (all instances of $class)
      */
     public function findInstanceOf($nodes, string $class) : array {
         return $this->find($nodes, function ($node) use ($class) {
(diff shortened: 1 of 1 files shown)


lib/PhpParser/NodeFinder.php
```diff
@@ -68,10 +70,12 @@ public function findFirst($nodes, callable $filter): ?Node {
...
-     * @param Node|Node[] $nodes  Single node or array of nodes to search in
-     * @param string      $class Class name
+     * @template TNode as Node
+     *
+     * @param Node|Node[] $nodes    Single node or array of nodes to search in
+     * @param class-string<TNode>   $class Class name
      *
-     * @return null|Node Found node, which is an instance of $class (or null if none found)
+     * @return null|TNode           Found node, which is an instance of $class (or null if none found)
      */
```
TomasVotruba, line 79: If needed, this can be separated to custo `@template-*` annotations per static analyzer, but I think it would only clutter the docblocks without any dadded value.

nikic:
Does this work in PhpStorm?

lib/PhpParser/NodeFinder.php
```diff
@@ -31,11 +31,13 @@ public function find($nodes, callable $filter) : array {
 
     /**
      * Find all nodes that are instances of a certain class.
+
+     * @template TNode as Node
      *
-     * @param Node|Node[] $nodes Single node or array of nodes to search in
-     * @param string      $class Class name
+     * @param Node|Node[] $nodes    Single node or array of nodes to search in
+     * @param class-string<TNode>   $class Class name
```
nikic, line 38: There are two many spaces

TomasVotruba:
PHPStorm behavior remains the same. It still sees the node as `Node`:

Peek 2022-08-08 21-51

lib/PhpParser/NodeFinder.php
TomasVotruba: Done :+1:

TomasVotruba:
:partying_face:

PR #876, opened 2022-08-28
Expr\Array_ can no longer contain null as item

ondrejmirtes:
After https://github.com/nikic/PHP-Parser/commit/68fc1ba4cb7b0aef9decc4e9d8f3b61eecdd3883 it's no longer possible to have `null` in this array.

### DIFF
--- a/lib/PhpParser/Node/Expr/Array_.php
+++ b/lib/PhpParser/Node/Expr/Array_.php
@@ -10,13 +10,13 @@ class Array_ extends Expr
     public const KIND_LONG = 1;  // array() syntax
     public const KIND_SHORT = 2; // [] syntax
 
-    /** @var (ArrayItem|null)[] Items */
+    /** @var ArrayItem[] Items */
     public $items;
 
     /**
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Oh, I noticed it can still be null before some post-processing in `fixupArrayDestructuring`.

Would be nice to have it eliminated because AST consumers will no longer see `null` there.

Do what you want with this PR :) Thanks.

nikic:
The general idea here makes sense, but it will take a bit more work, because we actually need to replace those null values with `Error` nodes. Currently the parser is happy to accept `[$x, , $y]` outside destructuring as well.

ondrejmirtes:
Alright, that's outside my skillset, I'll stay tuned :) Thank you.

ondrejmirtes:
Don't worry about it too much. Maybe it'd be sufficient to remove the null type as per this PR and be fine that it's slightly wrong for PHP-Parser internals?

ondrejmirtes:
Thank you very much!

PR #878, opened 2022-08-30
Extends Token tests

WinterSilence:
Copy from https://github.com/symfony/polyfill/blob/main/tests/Php80/PhpTokenTest.php


WinterSilence:
@nikic

PR #879, opened 2022-08-30
Fix Token methods

WinterSilence:
Fix result in cases when `pos` and/or `line` not set

### DIFF
--- a/lib/PhpParser/Token.php
+++ b/lib/PhpParser/Token.php
@@ -8,11 +8,17 @@ namespace PhpParser;
 class Token extends Internal\TokenPolyfill {
     /** Get (exclusive) zero-based end position of the token. */
     public function getEndPos(): int {
-        return $this->pos + \strlen($this->text);
+        return $this->pos < 0 ? -1 : $this->pos + \strlen($this->text);
     }
 
     /** Get 1-based end line number of the token. */
     public function getEndLine(): int {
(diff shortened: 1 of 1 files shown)


lib/PhpParser/Token.php
```diff
@@ -8,11 +8,14 @@
 class Token extends Internal\TokenPolyfill {
     /** Get (exclusive) zero-based end position of the token. */
     public function getEndPos(): int {
-        return $this->pos + \strlen($this->text);
+        return $this->pos < 0 ? $this->pos : $this->pos + \strlen($this->text);
```
nikic, line 11: Hm, I wonder if it wouldn't be better to throw an exception instead, if no end position/line is available.

WinterSilence: @nikic i'm lazy) this case need rewrites many tests (:

lib/PhpParser/Token.php
```diff
@@ -8,11 +8,14 @@
...
     }
 
     /** Get 1-based end line number of the token. */
     public function getEndLine(): int {
+        if ($this->line < 1 || !isset($this->text[1])) {
```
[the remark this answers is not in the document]
WinterSilence: @nikic to detect tokens contained single char

WinterSilence:
@nikic do you have any questions? why not pr not required yet?

PR #880, opened 2022-09-03
[format-preserving] Add test for adding attribute breaks method spacing

TomasVotruba:
Hi, we get reports in @rectorphp that adding attributes breaks the formatting of inner elements: 
https://github.com/rectorphp/rector/issues/7062

I think the issue is that original tokens do not match the new tokens position. The new tokens are added from the top, so by definition, anything bellow will be broken. Which is a pity, because people report huge classes/methods formatting removed.

There could some kind of token shift, based on newly added token count from the top. 

I looked into token printer, got a suspection about this area: https://github.com/nikic/PHP-Parser/blob/0201a7ee3f86c856eb009e1393fdd654b9667684/lib/PhpParser/PrettyPrinterAbstract.php#L564-L576
But after few trial I'm not sure how to shift tokens.

At least I'm sharing the test fixture, to verify the state. Any help/fix is appreciated :pray:

### DIFF
--- a/test/code/formatPreservation/adding_attribute_body_space.test
+++ b/test/code/formatPreservation/adding_attribute_body_space.test
@@ -1,0 +1,25 @@
+Add attribute body space
+-----
+<?php
+
+function some()
+{
+    $value = 100;
+
+    return $value;
+}
+
+-----
+$stmts[0]->attrGroups[] = new Node\AttributeGroup([
+    new Node\Attribute(new Node\Name('B'), [])
+]);
+-----
+<?php
+
+#[B]
+function some()
+{
+    $value = 100;
+
+    return $value;
+}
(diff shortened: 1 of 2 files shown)


nikic:
Should be fixed by https://github.com/nikic/PHP-Parser/commit/a2608f0b742b711d24de914d41c0162e37494ea6.

TomasVotruba:
Wow, thanks :+1: :

Can we add this test case, just to confirm for class methods?

TomasVotruba:
I've added test case directly to @rectorphp and it works well :+1:  Thank you. This gives us a big help with all attribute bug-reports   :bow: 

https://github.com/rectorphp/rector-src/commit/754fdd174ef743ebe7b528fc89adc3e61df7379e

PR #881, opened 2022-09-04
Fix inlined add multiple attribute group

samsonasik:
Based on https://github.com/nikic/PHP-Parser/commit/a2608f0b742b711d24de914d41c0162e37494ea6#commitcomment-82995459, the commit cause multiple attribute groups is inlined instead of multiple line. 

This PR try to fix it.

### DIFF
--- a/lib/PhpParser/PrettyPrinterAbstract.php
+++ b/lib/PhpParser/PrettyPrinterAbstract.php
@@ -925,9 +925,17 @@ abstract class PrettyPrinterAbstract
                     $result .= $insertStr;
                 }
                 $result .= $this->p($delayedAddNode, true);
+
+                if ($extraRight === "\n") {
+                    $result .= $this->nl ;
+                }
+
                 $first = false;
             }
-            $result .= $extraRight === "\n" ? $this->nl : $extraRight;
+
+            if ($extraRight !== "\n") {
+                $result .= $extraRight;
+            }
         }
 
         return $result;
--- a/test/code/formatPreservation/attributes.test
+++ b/test/code/formatPreservation/attributes.test
@@ -105,7 +105,11 @@ fn()
 $attrGroup = new Node\AttributeGroup([
     new Node\Attribute(new Node\Name('A'), []),
 ]);
+$attrGroup2 = new Node\AttributeGroup([
+    new Node\Attribute(new Node\Name('B'), []),
+]);
 $stmts[0]->attrGroups[] = $attrGroup;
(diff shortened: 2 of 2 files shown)


samsonasik:
Fixed 🎉

nikic:
I committed a slightly different fix in https://github.com/nikic/PHP-Parser/commit/44c6a97705a5611c6c2f3b4f20105bcb1aab666b. The missing newline here comes from delayed list insertion rather than empty list insertion, though in this particular case it ends up being the same.

samsonasik:
@nikic on rector, it seems make first AttributeGroup add new space:

```diff

There was 1 failure:

1) Rector\Doctrine\Tests\Set\DoctrineORM29Set\DoctrineORM29SetTest::test with data set #7 ('/home/runner/work/rector-src/...hp.inc')
Failed asserting that string matches format description.
--- Expected
+++ Actual
@@ @@
 #[Entity]
 class User
 {
-    #[ManyToOne(targetEntity: 'HelloWorld\UserGroup', inversedBy: 'users')]
+     #[ManyToOne(targetEntity: 'HelloWorld\UserGroup', inversedBy: 'users')]
     #[JoinColumn(name: 'group_id', referencedColumnName: 'id')]
```

ref https://github.com/rectorphp/rector-src/runs/8174961340?check_suite_focus=true#step:7:28

and somehow make `Invalid identifier name "0"`: 

```bash
 There were 4 errors:

1) Rector\Tests\DowngradePhp71\Rector\ConstFetch\DowngradePhp71JsonConstRector\DowngradePhp71JsonConstRectorTest::test with data set #3 ('/home/runner/work/rector-src/...hp.inc')
PHPStan\BetterReflection\Identifier\Exception\InvalidIdentifierName: Invalid identifier name "0"
```

 on different test , ref https://github.com/rectorphp/rector-src/runs/8174961410?check_suite_focus=true#step:7:28

samsonasik:
For reference, on rector-doctrine error, seems due to transformation from annotation to Attribute goes start from the annotation starter, so I create PR on rector side to change annotation indentation instead https://github.com/rectorphp/rector-doctrine/pull/134

nikic:
Just to double check, everything is fine now on the PHP-Parser side?

samsonasik:
Everything is fine, just unsure why replace annotation with attribute now start with the indent of annotation, not next stmt indentation, so:

```php
       /*
* @someAnnotationnn
*/
public $property;
```

now got:

```php
         #[someAnnotationnn]
public $property;
```

previously, it will make same indent

Issue #201, opened 2015-05-13
Inline comments with # getting associated with the wrong source-code element?


mindplay-dk:
I seem to be running into a parser issue, but can't be sure if this is a bug or as-intended. It seems buggy.

Code:

```PHP
$code = '<?php

$foo = 1;

$foo !== 2; # just because

$bar = 2;
';

$nodes = $parser->parse($code);

var_dump($nodes);
```

Output:

```
array(3) {
  [0] =>
  class PhpParser\Node\Expr\Assign#10 (4) {
    public $var =>
    class PhpParser\Node\Expr\Variable#8 (3) {
      public $name =>
      string(3) "foo"
      private $subNodeNames =>
      NULL
      protected $attributes =>
      array(2) {
        'startLine' =>
        int(3)
        'endLine' =>
        int(3)
      }
    }
    public $expr =>
    class PhpParser\Node\Scalar\LNumber#9 (3) {
      public $value =>
      int(1)
      private $subNodeNames =>
      NULL
      protected $attributes =>
      array(2) {
        'startLine' =>
        int(3)
        'endLine' =>
        int(3)
      }
    }
    private $subNodeNames =>
    NULL
    protected $attributes =>
    array(2) {
      'startLine' =>
      int(3)
      'endLine' =>
      int(3)
    }
  }
  [1] =>
  class PhpParser\Node\Expr\BinaryOp\NotIdentical#13 (4) {
    public $left =>
    class PhpParser\Node\Expr\Variable#11 (3) {
      public $name =>
      string(3) "foo"
      private $subNodeNames =>
      NULL
      protected $attributes =>
      array(2) {
        'startLine' =>
        int(5)
        'endLine' =>
        int(5)
      }
    }
    public $right =>
    class PhpParser\Node\Scalar\LNumber#12 (3) {
      public $value =>
      int(2)
      private $subNodeNames =>
      NULL
      protected $attributes =>
      array(2) {
        'startLine' =>
        int(5)
        'endLine' =>
        int(5)
      }
    }
    private $subNodeNames =>
    NULL
    protected $attributes =>
    array(2) {
      'startLine' =>
      int(5)
      'endLine' =>
      int(5)
    }
  }
  [2] =>
  class PhpParser\Node\Expr\Assign#17 (4) {
    public $var =>
    class PhpParser\Node\Expr\Variable#15 (3) {
      public $name =>
      string(3) "bar"
      private $subNodeNames =>
      NULL
      protected $attributes =>
[...]

PR #887, opened 2022-09-05
[tests] add test where comment is not joined with the class, if with attribute

TomasVotruba:
I tried this on Rector locally, and there is no docblock associated with the `Class_` if there is an attribute above.
Yet, the `Class_` node sees the attributes.

How can we reach it?

Reported in https://github.com/rectorphp/rector/issues/7453

### DIFF
--- a/test/code/formatPreservation/class_with_attibute_then_annotation.test
+++ b/test/code/formatPreservation/class_with_attibute_then_annotation.test
@@ -1,0 +1,23 @@
+Class with attribute then annotation
+-----
+<?php
+
+#[SomeAttribute]
+/**
+ * @SomeAnnotation
+ */
+final class SomeClass
+{
+}
+-----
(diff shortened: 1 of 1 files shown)


nikic:
The test is misleading in that this has nothing to do with formatting preservation. PHP-Parser simply does not support docblocks after attributes. Docblocks must be before attributes.

TomasVotruba:
I see. So this is a feature then?

nikic:
I wouldn't call it a feature, but I'm not sure it's worth adding support for this either. This is basically the same situation as writing something like this:
```
public function
/** Doc comment */
foobar() {}
```
The doc block is not in the correct position, and as such is not recognized. The situation with attributes is a bit different in that putting doc block between attributes and class is less obviously incorrect.

TomasVotruba:
I see. I though about the same way, also in context of https://github.com/nikic/PHP-Parser/issues/201

Thanks for confirmation, I'll recommend to use a e.g. PHP CS Fixer to handle these flips.

PR #889, opened 2022-09-08
Fixed type in `UnionType`

kukulich:


lib/PhpParser/Node/IntersectionType.php
```diff
@@ -6,13 +6,13 @@
 
 class IntersectionType extends ComplexType
 {
-    /** @var (Identifier|Name)[] Types */
+    /** @var (Identifier|Name|UnionType)[] Types */
```
nikic, line 9: I don't think this is right. `(X&Y)|Z` is allowed, but `(X|Y)&Z` is not.

kukulich: You are right. Fixed.

Issue #662, opened 2020-05-12
numberSeparators.test fails on armhf


bryceharrington:
In the Ubuntu CI, the new php-parser 4.4.0 fails on one of the newly added test cases, only on armhf, with this error:

```
1) PhpParser\CodeParsingTest::testParse with data set "tmp/autopkgtest.4zawvj/build.7wd/src/test/code/parser/scalar/numberSeparators.test#0" ('Different integer syntaxes (/....test)', '<?php\n\n6.674_083e-11;\n299_...n1e_2;', 'Syntax error, unexpected T_ST...  )\n)', null)
Different integer syntaxes (/tmp/autopkgtest.4zawvj/build.7wd/src/test/code/parser/scalar/numberSeparators.test)
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
         )
     )
     2: Stmt_Expression(
-        expr: Scalar_LNumber(
-            value: 3405705229
+        expr: Scalar_DNumber(
+            value: 0
         )
     )

```
I'm guessing this is checking range on long data types, maybe a signed/unsigned mismatch?  For ARM, it looks like a signed long is 32 bits word-aligned, or -2,147,483,648 to 2,147,483,647, and guessing 3,405,705,229 is getting returned as 0 as it's out of range?

https://objectstorage.prodstack4-5.canonical.com/v1/AUTH_77e2ada1e7a84929a74ba3b87153c0ac/autopkgtest-groovy/groovy/armhf/p/php-parser/20200501_165040_5a5c8@/log.gz

Downstream bug:  https://bugs.launchpad.net/ubuntu/+source/php-parser/+bug/1878102

bryceharrington:
There is a similar, related test failure:



2) PhpParser\Lexer\EmulativeTest::testLexNewFeatures with data set #20 ('0xCAFE_F00D', array(array(312, '0xCAFE_F00D')))
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
 Array &0 (
     0 => Array &1 (
- 0 => 312
+ 0 => 313
         1 => '0xCAFE_F00D'
     )
 )

'0xCAFE_F00D' is hexadecimal for 3405705229, which is the same number as in the syntax issue above. The test is receiving this as a T_DNUMBER (i.e. type number 313) rather than as a T_LNUMBER (type 312) as defined in Tokens.php.

nikic:
[...]

PR #896, opened 2022-11-12
Skip incompatible armhf tests

athos-ribeiro:
As pointed out in #662, there are a couple of tests failing for armhf.Possible alternatives to fix the issue include skipping the failing tests in arm or reducing the size of the integers being tested. This PR implements the former approach to kick-off discussions on the matter.

P.S.: I could not find a way of introducing an armhf CI here without providing a self hosted infrastructure.

### DIFF
--- a/test/PhpParser/CodeParsingTest.php
+++ b/test/PhpParser/CodeParsingTest.php
@@ -11,6 +11,10 @@ class CodeParsingTest extends CodeTestAbstract
      * @dataProvider provideTestParse
      */
     public function testParse($name, $code, $expected, $modeLine) {
+        if (strpos(php_uname('m'), 'arm') !== false and strpos($name, '/numberSeparators.test')) {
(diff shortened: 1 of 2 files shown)


nikic:
Thanks for the PR! I looked into this in more detail, and it turns out that this exposed an implementation bug in parsing of overflowing hex literals, which is fixed in https://github.com/nikic/PHP-Parser/commit/7027899d7f428dcc108aaaada8cc1ab2e4672cc7. For the failing tests themselves, I've slightly tweaked them to stay as integers on both 32-bit and 64-bit in https://github.com/nikic/PHP-Parser/commit/e072fd2c30e506516dd4677b8e1d6142609e1c37 -- we already have separate tests that overflow on both 32-bit and 64-bit.

athos-ribeiro:
Hi Nikita,

Thanks for looking into this one! Sorry for not being more helpful with the PR :)

PR #897, opened 2022-11-13
added type int to argument of addAlias function

AlessandroMinoccheri:
The type of argument `type` is always `int` so I added it.


nikic:
FWIW, this was already present on the master branch.

Issue #899, opened 2022-11-26
Sometimes it makes sense for visitors to leave in reverse order of entering


rulatir:
Speaking from my general experience with using the vistor pattern, I find that sometimes certain traversal scenarios are easier to engineer if the traverser calls `->leave()` in reverse order of visitors, i.e. first enter last leave.

Two questions:

 1. Do you believe implementing this would be as simple as an optional `array_reverse()` in `traverseNode()`, and it won't cause problems like breaking the shipped visitors (node connectors, finders etc.)?
 2. If you believe it should be feasible, would you accept a PR that implements this and adds the necessary knob to `NodeTraverser`?

nikic:
Reverse order for leave makes sense to me -- possibly we can just implement that change on the master branch, without an option? The current behavior is arguably "incorrect", but probably also not something that should be changed in a point release.

theofidry:
FWIW if you provide a change on a branch / PR I can give it a go at PHP-Scoper and maybe could be tested on PHPStan & Rector. If nothing breaks it's a good sign, otherwise it means it will need a good warning

dryabov:
Try this patch: https://github.com/nikic/PHP-Parser/pull/901

theofidry:
Thanks! will give it a go ASAP

theofidry:
Sorry was quite busy with other stuff and then forgot about it. I'll give it a go tonight otherwise it will have to wait after Christmas for me to try it on PHP-Scoper

theofidry:
All green on PHP-Scoper side. But I think @TomasVotruba who is also using it heavily with a few exotic usages in Rector may want to have a go at it

dryabov:
Just a side note: maybe it would be better (for maintenance) to introduce a `ReverseIterator` class and rewrite the patch via
```php
foreach (new ReverseIterator($this->visitors) as $visitorIndex => $visitor) {
```
instead of direct `end`/`prev` manipulations.

The `ReverseIterator` class can be easily implemented in the following way:
```php
class ReverseIterator implements Iterator
{
    public function __construct(
        private iterable $iterator
    ) {}
    public function rewind(): void {
        end($this->iterator);
    }
    public function valid(): bool {
        return key($this->iterator) !== null;
    }
    public function current(): mixed {
        return current($this->iterator);
    }
    public function key(): mixed {
        return key($this->iterator);
    }
    public function next(): void {
        prev($this->iterator);
    }
}
```

rulatir:
[...]

PR #901, opened 2022-12-05
Fix issue #899

dryabov:
A patch to call `leaveNode` in the reverse order.


nikic:
I've implemented a variant of this patch in https://github.com/nikic/PHP-Parser/commit/8490c0e82d8ee09e37a2e426af773da9661cbd29. I'm not sure why CI did not run on this commit, as that would have shown test failures :/

PR #906, opened 2022-12-14
rewrote overly magic code to make it readable 

staabm:


lib/PhpParser/Builder/Class_.php
```diff
@@ -123,8 +123,6 @@ public function addStmt($stmt) {
             throw new \LogicException(sprintf('Unexpected node of type "%s"', $stmt->getType()));
         }
 
-        $targets[$class][] = $stmt;
```
nikic, line 126: This is not dead code: `$targets[$class]` is a reference to a property, so this modifies that property.

That said, this code is unnecessarily clever, and if you want to rewrite it like this instead, I wouldn't mind: https://github.com/nikic/PHP-Parser/blob/950bf8f1d1865cfd0d248e13a3455db0be1d870c/lib/PhpParser/Builder/Trait_.php#L43

staabm: just pushed a new commit with more readable code.

theofidry:
Holy crap only now I understand what that code was doing 😅 

Quite clever indeed

Issue #910, opened 2023-01-13
Syntax error with readonly and DNF tokens


mrsuh:
How to reproduce:
```bash
composer req nikic/php-parser:v4.15.2
```

test.php
```php
<?php

require_once __DIR__ . '/vendor/autoload.php';

use PhpParser\ParserFactory;

$code = <<<'CODE'
<?php

class A {}
class B {}

class Container  {
    public readonly (A&B)|null $property; // <-- error
}
CODE;

$parser = (new ParserFactory)->create(ParserFactory::PREFER_PHP7);
$parser->parse($code);
```

This error is only reproducible with PHP < 8.2
```bash
docker run -it --rm -v "$PWD":/app php:7.4-cli php /app/test.php
docker run -it --rm -v "$PWD":/app php:8.0-cli php /app/test.php
docker run -it --rm -v "$PWD":/app php:8.1-cli php /app/test.php
```

```bash
Fatal error: Uncaught PhpParser\Error: Syntax error, unexpected '(', expecting T_VARIABLE on line 7 in /app/vendor/nikic/php-parser/lib/PhpParser/ParserAbstract.php:318
Stack trace:
#0 /app/vendor/nikic/php-parser/lib/PhpParser/ParserAbstract.php(159): PhpParser\ParserAbstract->doParse()
#1 /app/test.php(28): PhpParser\ParserAbstract->parse('<?php\n\nclass A ...')
#2 {main}
  thrown in /app/vendor/nikic/php-parser/lib/PhpParser/ParserAbstract.php on line 318
```

I fixed error for PHP < 8.1 like [this](https://github.com/mrsuh/PHP-Parser/blob/4d76bd892b26b158d0644bcd4468c18f0cf0ab02/lib/PhpParser/Lexer/TokenEmulator/ReadonlyTokenEmulator.php#L30-L41).

Error for PHP 8.1 is more complicated because of function `token_get_all()`.
How to reproduce `token_get_all()` error for PHP 8.1:

tokens.php
```php
<?php

$code = <<<'CODE'
<?php

class A {}
class B {}

class Container  {
    public readonly (A&B)|null $property1;
    public readonly null|(A&B) $property2;
}
CODE;

foreach (token_get_all($code) as $token) {
    if (is_array($token)) {
        echo "Line {$token[2]}: ", token_name($token[0]), " ('{$token[1]}')", PHP_EOL;
    }
}
```

```bash
[...]

PR #911, opened 2023-01-13
Fix issue #910

mrsuh:


nikic:
I've fixed this in a different way in https://github.com/nikic/PHP-Parser/commit/8863f92b58119e4093437feeef65714b9f922d54. The emulation needs to match whatever PHP produces in the latest version, so we have to always produce T_READONLY, even though it is annoying here.

PR #924, opened 2023-06-04
Early return false after VariadicPlaceholder check on CallLike::isFirstClassCallable()

samsonasik:
If I understand correctly, `VariadicPlaceholder` will only exists on first arg, after it, it will always an Arg. so no need whole loop to verify another `VariadicPlaceholder`, this can speed up usage on `getArgs()` method as assert check.


nikic:
Yes, at least currently the variadic placeholder will be the one and only argument, at least for nodes produced by the parser.

So possibly it makes sense to check for `\count() === 1` and only check that one argument rather than a loop that always breaks after one iteration?

samsonasik:
Ok, I updated to use count() equal to 1 and check current arg is a VariadicPlaceholder instance.

PR #932, opened 2023-07-22
Option to keep redundant parentheses

theriault:
I know this goes against a true AST, but could there be an option to keep redundant parentheses?

Something similar to https://clang.llvm.org/doxygen/classclang_1_1ParenExpr.html#details

### DIFF
--- a/grammar/php.y
+++ b/grammar/php.y
@@ -1021,7 +1021,7 @@ expr:
     | expr '>' expr                                         { $$ = Expr\BinaryOp\Greater       [$1, $3]; }
     | expr T_IS_GREATER_OR_EQUAL expr                       { $$ = Expr\BinaryOp\GreaterOrEqual[$1, $3]; }
     | expr T_INSTANCEOF class_name_reference                { $$ = Expr\Instanceof_[$1, $3]; }
-    | '(' expr ')'                                          { $$ = $2; }
+    | '(' expr ')'                                          { $$ = $this->keepRedundantParentheses ? Expr\Parenthesized[$2] : $2; }
     | expr '?' expr ':' expr                                { $$ = Expr\Ternary[$1, $3,   $5]; }
     | expr '?' ':' expr                                     { $$ = Expr\Ternary[$1, null, $4]; }
     | expr T_COALESCE expr                                  { $$ = Expr\BinaryOp\Coalesce[$1, $3]; }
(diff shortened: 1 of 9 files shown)


nikic:
I'm generally open to the idea -- what do you want to use this for?

theriault:
@nikic I am working on an automated refactor of a legacy codebase where files contain a mix of functions and side-effects. Each file will become a class of static functions, where the side-effects will become the body of a "catchall" static function and the other functions becoming a part of the class; rescoping all the code.

Preserving the existing spacing/comments while simultaneously  indenting/relocating large chunks of code is not working well with the existing preservation feature of  `PrettyPrinter`, so I am working on a `before/after` comment/spacing consumer `NodeVisitor` and a special `PrettyPrinter` that then utilizes the `before/after` attributes, so that I can move chunks of code around more easily without worrying too much about the source comments/spacing getting mangled/lost as each `Node` can define surrounding spacing/comments (and if those attributes aren't defined, the default formatting kicks in). Each comment/spacing token will be flagged as being owned by only one node for existing code.

To do this, I need to try to maintain location information of redundant parentheses and such so that surrounding comments/spacing can be demarcated/consumed by what may ultimately be redundant parentheses, e.g.

```php
if (
     (/*some condition*/) // some comment around this condition 
     || (/* some other condition */) // some comment around this other condition
) {
   //...
}
```

Ideally, all parenthesized and braced chunks of code could be captured, so even if a dev did some outlandish style like the following, it could be captured.

```php
$something = 123;

// commenting/grouping a chunk of code
{
       // code within its own redundant block
}

// commenting/grouping another chunk of code
{
       // code within its own redundant block
}
```

However, this PR in its current state only handles redundant parentheses as I haven't seen the above style too much outside of intermediary refactors of code or code that required an anonymous scope (which PHP doesn't support).

theriault:
Closing due to staleness

nikic:
Sorry for the long silence here. I think this is a reasonable feature, with little cost. A nice side benefit would be that you could easily get some more explicit control over pretty printing, which goes for a "least parentheses" approach (when not preserving formatting), which is not always desirable.

I looked over the implementation and it looks good to me, just needs a rebase.

nikic:
> However, this PR in its current state only handles redundant parentheses as I haven't seen the above style too much outside of intermediary refactors of code or code that required an anonymous scope (which PHP doesn't support).

By the way, PHP-Parser 5 does create `Stmt\Block` nodes for these kinds of redundant blocks. (Likely I added that after your comment was written...)

PR #935, opened 2023-07-30
Backport PHP 8.3 support to PHP-Parser 4.x

ondrejmirtes:
Hello,
PHPStan is a bit between a rock and a hard place regarding PHP 8.3 support. I understand you want to support PHP 8.3 features only with PHP-Parser 5.

But I can upgrade to PHP-Parser 5 only in PHPStan 2.0. Additionally, PHPStan depends on BetterReflection which also needs to release a new version that works with PHP-Parser 5 before I can use it in PHPStan. And also Rector depends on PHPStan which means that Rector can support PHP 8.3 only once PHPStan does.

But I want to support PHP 8.3 as soon as possible with PHPStan and I don't want to be forced to release PHPStan 2.0 sooner than I'd want because of that. I'd like to have PHP 8.3 support and 2.0 release as separate milestones.

So I'd like to decouple and deconstruct this situation to remove a bottleneck and to make this easier between PHPStan <-> BetterReflection <-> Rector. Support for PHP 8.3 in 4.x would really help me with that.

I prepared a branch where I cherry-picked the commits. I'd understand if you don't want to merge this. In that case I'll use a temporary PHP-Parser fork in PHPStan 1.x with these commits. (Rector will also likely need to use the same fork.)

Thanks for understanding.

/cc @TomasVotruba


lib/PhpParser/Internal/PrintableNewAnonClassNode.php
```diff
@@ -19,6 +19,8 @@ class PrintableNewAnonClassNode extends Expr
 {
     /** @var Node\AttributeGroup[] PHP attribute groups */
     public $attrGroups;
+	/** @var int Modifiers */
+	public $flags;
```
nikic, line 23: Broken indent

ondrejmirtes: Fixed

ondrejmirtes:
Thank you, I really appreciate it!

PR #940, opened 2023-08-15
Normalize enum value to ClassConstFetch

### LINKED ISSUE #930: `Enum` as a `Attribute` parameter
Hello!

I think I found bug. Lets consider following snippet:

```php
<?php

use PhpParser\BuilderFactory;
use PhpParser\Node\Name;
use PhpParser\PrettyPrinter\Standard;

enum HttpMethod: string
{
    case GET = 'GET';
}

#[Attribute(Attribute::TARGET_METHOD)]
readonly class HTTP
{
    public HttpMethod $httpMethod;

    public function __construct(
        HttpMethod|string $httpMethod,
        public string $path
    )
    {
        if (is_string($httpMethod)) {
            $httpMethod = HttpMethod::from($httpMethod);
        }
        $this->httpMethod = $httpMethod;
    }
}

interface Api
{
    #[HTTP(httpMethod: HttpMethod::GET, path: '/users')]
    public function getUsers();
}

$reflectionClass = new ReflectionClass(Api::class);

$builderFactory = new BuilderFactory();

$class = $builderFactory->class('ApiImpl')
    ->implement('\\' . Api::class);

$reflectionMethod = $reflectionClass->getMethod('getUsers');

$method = $builderFactory
    ->method($reflectionMethod->getName())
    ->makePublic();

$reflectionAttributes = $reflectionMethod->getAttributes();
foreach ($reflectionAttributes as $reflectionAttribute) {
    $name = new Name($reflectionAttribute->getName());
    $attribute = $builderFactory->attribute($name, $reflectionAttribute->getArguments());

    $method->addAttribute($attribute);
}

$class->addStmt($method->getNode());

$proxyNamespaceBuilder = $builderFactory
    ->namespace('Namespace')
    ->addStmt($class);

$standard = new Standard();
echo $standard->prettyPrint([$proxyNamespaceBuilder->getNode()]);
```

When I try to create a proxy class with method which takes `enum` as a parameters I receive an exception:

```
Fatal error: Uncaught LogicException: Invalid value in /opt/project/vendor/nikic/php-parser/lib/PhpParser/BuilderHelpers.php on line 280

Call Stack:
    0.0004     401096   1. {main}() /opt/scratches/scratch_37.php:0
    0.0052    1342792   2. PhpParser\BuilderFactory->attribute($name = class PhpParser\Node\Name { protected $attributes = []; public $parts = [0 => 'HTTP'] }, $args = ['httpMethod' => enum HttpMethod::GET('GET'), 'path' => '/users']) /opt/scratches/scratch_37.php:54
    0.0052    1346616   3. PhpParser\BuilderFactory->args($args = ['httpMethod' => enum HttpMethod::GET('GET'), 'path' => '/users']) /opt/project/vendor/nikic/php-parser/lib/PhpParser/BuilderFactory.php:26
    0.0053    1351336   4. PhpParser\BuilderHelpers::normalizeValue($value = enum HttpMethod::GET('GET')) /opt/project/vendor/nikic/php-parser/lib/PhpParser/BuilderFactory.php:252
```

When I added a following handling in `BuilderHelpers::normalizeValue()` code works OK.

```php
public static function normalizeValue($value) : Expr {
    if ($value instanceof Node\Expr) {
        return $value;
    }

    if (is_null($value)) {
        return new Expr\ConstFetch(
            new Name('null')
        );
    }

    if (is_bool($value)) {
        return new Expr\ConstFetch(
            new Name($value ? 'true' : 'false')
        );
    }

    if (is_int($value)) {
        return new Scalar\LNumber($value);
    }

    if (is_float($value)) {
        return new Scalar\DNumber($value);
    }

    if (is_string($value)) {
        return new Scalar\String_($value);
    }

    if (is_array($value)) {
        $items = [];
        $lastKey = -1;
        foreach ($value as $itemKey => $itemValue) {
            // for consecutive, numeric keys don't generate keys
            if (null !== $lastKey && ++$lastKey === $itemKey) {
                $items[] = new Expr\ArrayItem(
                    self::normalizeValue($itemValue)
                );
            } else {
                $lastKey = null;
                $items[] = new Expr\ArrayItem(
                    self::normalizeValue($itemValue),
                    self::normalizeValue($itemKey)
                );
            }
        }

        return new Expr\Array_($items);
    }

    // !!! HERE !!!
    if (is_object($value)) {
        return new Expr\ClassConstFetch(new Name($value::class), $value->name);
    }

    throw new \LogicException('Invalid value');
}
```

What do you think about this fix? It's ok? Or it should be handled in the another way?

ruudk:
Fixes #930

### DIFF
--- a/lib/PhpParser/BuilderHelpers.php
+++ b/lib/PhpParser/BuilderHelpers.php
@@ -6,6 +6,7 @@ use PhpParser\Node\ComplexType;
 use PhpParser\Node\Expr;
 use PhpParser\Node\Identifier;
 use PhpParser\Node\Name;
+use PhpParser\Node\Name\FullyQualified;
 use PhpParser\Node\NullableType;
 use PhpParser\Node\Scalar;
 use PhpParser\Node\Stmt;
@@ -215,7 +216,7 @@ final class BuilderHelpers
      * Normalizes a value: Converts nulls, booleans, integers,
      * floats, strings and arrays into their respective nodes
      *
-     * @param Node\Expr|bool|null|int|float|string|array $value The value to normalize
+     * @param Node\Expr|bool|null|int|float|string|array|\UnitEnum $value The value to normalize
      *
      * @return Expr The normalized value
      */
@@ -267,6 +268,10 @@ final class BuilderHelpers
             }
 
             return new Expr\Array_($items);
+        }
+
+        if ($value instanceof \UnitEnum) {
+            return new Expr\ClassConstFetch(new FullyQualified(\get_class($value)), new Identifier($value->name));
         }
 
         throw new \LogicException('Invalid value');
--- a/lib/PhpParser/Builder/ClassConst.php
+++ b/lib/PhpParser/Builder/ClassConst.php
@@ -26,7 +26,7 @@ class ClassConst implements PhpParser\Builder
      * Creates a class constant builder
      *
      * @param string|Identifier                          $name  Name
-     * @param Node\Expr|bool|null|int|float|string|array $value Value
+     * @param Node\Expr|bool|null|int|float|string|array|\UnitEnum $value Value
      */
     public function __construct($name, $value) {
         $this->constants = [new Const_($name, BuilderHelpers::normalizeValue($value))];
@@ -36,7 +36,7 @@ class ClassConst implements PhpParser\Builder
      * Add another constant to const group
      *
      * @param string|Identifier                          $name  Name
-     * @param Node\Expr|bool|null|int|float|string|array $value Value
+     * @param Node\Expr|bool|null|int|float|string|array|\UnitEnum $value Value
      *
      * @return $this The builder instance (for fluid interface)
      */
--- a/lib/PhpParser/BuilderFactory.php
+++ b/lib/PhpParser/BuilderFactory.php
@@ -213,7 +213,7 @@ class BuilderFactory
     /**
      * Creates node a for a literal value.
      *
-     * @param Expr|bool|null|int|float|string|array $value $value
+     * @param Expr|bool|null|int|float|string|array|\UnitEnum $value $value
      *
      * @return Expr
      */
--- a/test/PhpParser/BuilderHelpersTest.php
+++ b/test/PhpParser/BuilderHelpersTest.php
@@ -3,6 +3,8 @@
 namespace PhpParser;
 
 use PhpParser\Builder\Class_;
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Name\FullyQualified;
 use PhpParser\Node\Scalar;
 use PhpParser\Node\Stmt;
 use PhpParser\Node\Expr;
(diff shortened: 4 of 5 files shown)


test/PhpParser/BuilderHelpersPHP81Test.php
```diff
@@ -0,0 +1,19 @@
...
+use PhpParser\Node\Name\FullyQualified;
+
+class BuilderHelpersPHP81Test extends \PHPUnit\Framework\TestCase
+{
+    public function testNormalizeValueEnum() {
+        if (\PHP_VERSION_ID <= 80100) {
+            $this->markTestSkipped('Enums are supported since PHP 8.1');
+        }
+
+        include __DIR__ . '/../code/Suit.php';
```
ruudk, line 15: @nikic Is this the correct way to handle this?

vitek-rostislav:
Hi, just encountered the problem described in https://github.com/nikic/PHP-Parser/issues/930. I can confirm the solution proposed   here fixes the issue. What is the status of this PR, please? It has been pending for a while now

ruudk:
@nikic could you have a look please 🙏🏻

adrianpaiva1:
@ruudk @nikic any update on this? I'm having the same issue.

ruudk:
Waiting for a merge 😁

test/PhpParser/BuilderHelpersPHP81Test.php
nikic: This seems reasonable (alternative would be eval). But this shouldn't be inside the code/ directory.

lib/PhpParser/BuilderHelpers.php
```diff
@@ -215,7 +216,7 @@ public static function normalizeType($type) {
      * Normalizes a value: Converts nulls, booleans, integers,
      * floats, strings and arrays into their respective nodes
      *
-     * @param Node\Expr|bool|null|int|float|string|array $value The value to normalize
+     * @param Node\Expr|bool|null|int|float|string|array|\UnitEnum $value The value to normalize
```
nikic, line 219: This doc comment change also needs to be propagated to most users of normalizeValue().

test/PhpParser/BuilderHelpersPHP81Test.php
```diff
@@ -0,0 +1,20 @@
+<?php declare(strict_types=1);
+
+namespace PhpParser;
+
+use PhpParser\Node\Expr;
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Name\FullyQualified;
+
+class BuilderHelpersPHP81Test extends \PHPUnit\Framework\TestCase
```
nikic, line 9: Can merge this into the main BuilderHelpersTest, as the PHP version condition is inside the method anyway?

test/PhpParser/BuilderHelpersPHP81Test.php
ruudk: Created an `fixtures` directory and moved it there + rebased the PR.

lib/PhpParser/BuilderHelpers.php
ruudk: I checked all usages, but I only added it to `\PhpParser\BuilderFactory::val`

test/PhpParser/BuilderHelpersPHP81Test.php
ruudk: Done!

lib/PhpParser/BuilderHelpers.php
nikic: Why only that one? Enums are valid values for property/param/const defaults.

ruudk: Those mentioned are typed as `mixed`. 

But I found 2 more in `\PhpParser\Builder\ClassConst`: `__construct` and `addConst`.

ruudk: Done

nikic (approved):

PR #945, opened 2023-09-12
Use more precise Use_::TYPE_* types

staabm:
for better static analysis support in consuming projects


lib/PhpParser/Builder/Use_.php
```diff
@@ -10,7 +10,10 @@
 class Use_ implements Builder {
     /** @var Node\Name */
     protected Node\Name $name;
-    /** @var int */
+    /**
+     * @var int
+     * @phpstan-var Stmt\Use_::TYPE_*
```
nikic, line 15: I don't think we need `phpstan-` here, this seems to be generally supported.

staabm: good catch, fixed

PR #946, opened 2023-09-18
Fix missing type annotations in AST nodes

rainx:
# Description

After upgrading to v5.0.0, there were some missing type annotations in the AST nodes class which those types were correctly annotated in v4.x

# What does this PR do

Add the missing types

# Why did we find this issue? 

I am working on a project that generates TypeScript type definition files for the PHP Parser dumped AST nodes(by `-j` option of the  CLI command). those types should be generated automatically through the PHP source code, it depends on the type annotation to infer the proper TS types. 

It works very well for v4. today, I tried to migrate it to v5. and when I rerun the `generate` script. I found the missing types.

### DIFF
--- a/lib/PhpParser/Node/Expr/ArrowFunction.php
+++ b/lib/PhpParser/Node/Expr/ArrowFunction.php
@@ -7,8 +7,10 @@ use PhpParser\Node\Expr;
 use PhpParser\Node\FunctionLike;
 
 class ArrowFunction extends Expr implements FunctionLike {
+    /** @var bool Whether the closure is static */
     public bool $static;
 
+    /** @var bool Whether to return by reference */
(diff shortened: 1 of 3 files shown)


lib/PhpParser/Node/Expr/ArrowFunction.php
```diff
@@ -7,8 +7,10 @@
 use PhpParser\Node\FunctionLike;
 
 class ArrowFunction extends Expr implements FunctionLike {
+    /** @var bool */
```
nikic, line 10: You need to add descriptions to any redundant type annotations you want to preserve.

rainx: Just added them, please check it.

PR #952, opened 2023-10-03
Simple fix miss property...(class member)

xjaja:


nikic:
This looks like spam. Please explain what you're trying to do if it isn't.

grammar/php7.y
```diff
@@ -708,7 +708,7 @@ static_var:
 ;
 
 class_statement_list_ex:
-      class_statement_list_ex class_statement               { if ($2 !== null) { push($1, $2); } }
+      class_statement_list_ex class_statement               { if ($2 !== null) { push($1, $2); }; $$ = $1; }
```
nikic, line 711:
```suggestion
      class_statement_list_ex class_statement               { if ($2 !== null) { push($1, $2); } else { $$ = $1; } }
```

nikic (commented):
Okay, I understand what you are trying to do now. You need to run `grammar/rebuildParser.php` to regenerate the parser, otherwise this will have no effect.

nikic:
This still needs to run grammar/rebuildParser.php to regenerate the parser.

xjaja:
Thx.

PR #958, opened 2023-11-01
build: Exclude grammar from export git artifact

### LINKED ISSUE #957: Global symbols usage
Whilst working on [PHP-Scoper](https://github.com/humbug/php-scoper) and reviewing the symbols it is exposing*, I found that PHP-Parser had a few global symbols being used:

- `LIB`
- `PARAMS`
- `ARGS`
- `preprocessGrammar`
- `resolveNodes`
- ...

I am wondering if it is truly necessary for them to be registered as global symbols?

*: PHP-Scoper changes the namespace, and for not breaking globally registered user functions, it aliases the ones it came across with a statement like so:

```php
if (!function_exists('preprocessGrammar')) { function preprocessGrammar() { return \_HumbugBox57f37d5ac4d1\preprocessGrammar(...func_get_args()); } }
```

theofidry:
Closes #957.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,6 @@
 /.github         export-ignore
 /doc             export-ignore
+/grammar         export-ignore
 /test            export-ignore
 /test_old        export-ignore
 .editorconfig    export-ignore


theofidry:
Thank you @nikic!

Issue #986, opened 2024-03-17
php-parser v4.19.0 got deprecated Optional parameter before required parameter on PrintableNewAnonClassNode.php


samsonasik:
on php-parser v4.19.0, we tried using on rector-src, and got error:

```
PHP Deprecated:  Optional parameter $extends declared before required parameter $attributes is implicitly treated as a required parameter in /Users/samsonasik/www/rector-src/vendor/nikic/php-parser/lib/PhpParser/Internal/PrintableNewAnonClassNode.php on line 33
Deprecated: Optional parameter $extends declared before required parameter $attributes is implicitly treated as a required parameter in /Users/samsonasik/www/rector-src/vendor/nikic/php-parser/lib/PhpParser/Internal/PrintableNewAnonClassNode.php on line 33
```

samsonasik:
I created a patch on php-parser itself on:

- https://github.com/nikic/PHP-Parser/pull/987

PR #987, opened 2024-03-17
Fix deprecated Optional parameter before required parameter on PrintableNewAnonClassNode

samsonasik:
Fixes https://github.com/nikic/PHP-Parser/issues/986

### DIFF
--- a/lib/PhpParser/Internal/PrintableNewAnonClassNode.php
+++ b/lib/PhpParser/Internal/PrintableNewAnonClassNode.php
@@ -31,8 +31,8 @@ class PrintableNewAnonClassNode extends Expr
     public $stmts;
 
     public function __construct(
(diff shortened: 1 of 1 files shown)


samsonasik:
Ready to merge 👍

staabm:
Would this error be detected by a php lint CI job? 

If so, would be great to add such github actions

samsonasik:
On rector-sec, we use list of classes as preload, and error detected early.

I compared to php-parser v5, it seems `null` default value is not exists, so I think this is solution with remove default null.

staabm:
I don't question your fix. was only thinking about some CI tooling to make sure we can detect such errors in the future earlier

nikic:
I don't see the deprecation warning printed on the PHP 8.2 and PHP 8.3 CI jobs at all -- I wonder why that is. I do see it when running with these versions locally.

PR #989, opened 2024-03-17
Make phpunit fail on deprecation warnings

staabm:


staabm:
interessting.. locally I get phpunit fail instead of accepting the error. I think this means the deprecation warning is not emitted at all in CI

staabm:
ok, we can see CI failling now. will re-revert the initial fix now

staabm:
@nikic I think this fix should cover us in the future

samsonasik (approved):

.github/workflows/main.yml
```diff
@@ -49,6 +49,7 @@ jobs:
         with:
           coverage: "none"
           php-version: "${{ matrix.php-version }}"
+          ini-values: "error_reporting=2147483647" # E_ALL
```
nikic, line 52:
```suggestion
          ini-values: "error_reporting=E_ALL"
```
Does this not work?

staabm:
opened a setup-php bug https://github.com/shivammathur/setup-php/issues/827

phpunit.xml.dist
```diff
@@ -4,6 +4,10 @@
          xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
          backupGlobals="false"
          colors="true"
+         convertErrorsToExceptions="true"
+         convertNoticesToExceptions="true"
+         convertWarningsToExceptions="true"
+         convertDeprecationsToExceptions="true"
```
nikic, line 10: I think you only need this one, the rest default to true.

.github/workflows/main.yml
staabm: seems to work

staabm: using the fix [suggested by the setup-php author](https://github.com/shivammathur/setup-php/issues/827#issuecomment-2002367665) now

nikic:
Quite unexpected behavior to me ... guess it's time to go through all my repos and add `ini-file: "development"` to CI actions.

staabm:
> Quite unexpected behavior to me

agree. the author agrees to. its planned for the next major, because its a breaking change.

see https://github.com/shivammathur/setup-php/issues/827#issuecomment-2002367665

PR #995, opened 2024-04-19
Add getTypeNode and toString to Param for getting more infos

gseidel:
At the moment, it is not possible to have access to type and name information of a function parameter, because `Node\Param` give no access to read the `$type` or `$var` property. `getType` will just return the string `Param`. So I added a simple getter `getTypeNode` to provide type infos and `toString` to get the name of the parameter (Same pattern like in function).

This PR shouldn't have any BC breaks.

### DIFF
--- a/lib/PhpParser/Node/Param.php
+++ b/lib/PhpParser/Node/Param.php
@@ -60,6 +60,13 @@ class Param extends NodeAbstract {
     }
 
     /**
+     * @return null|Identifier|Name|ComplexType
+     */
+    public function getTypeNode(): ?Node {
+        return $this->type;
+    }
+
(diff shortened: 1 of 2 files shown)


staabm:
the property your new getter is returning is public. you can access the property directly.

gseidel:
@staabm Damn, thats right. Thanks for pointing this out. I guess, I working to much on projects where class properties are private or protected, so I didn't saw that as an option when I was working with it 😅

nikic:
Yeah, you're supposed to access properties directly :) The only reason why there are methods for some things is that PHP doesn't support properties in interfaces, so we have to go through methods in that case.

PR #997, opened 2024-05-13
Add missing void return types

jorgsowa:
Added missing `void` return type to the methods.


lib/PhpParser/NodeVisitor/NodeConnectingVisitor.php
```diff
@@ -25,12 +25,12 @@ final class NodeConnectingVisitor extends NodeVisitorAbstract {
      */
     private $previous;
 
-    public function beforeTraverse(array $nodes) {
+    public function beforeTraverse(array $nodes): void {
```
nikic, line 28: Should leave these methods alone, they aren't quite void from an interface perspective.

PR #998, opened 2024-05-13
Adjust tests to PHPUnit 10

jorgsowa:
Prepares codebase to PHPUnit 10.

- migrated PHPUnit configuration to v10 and left old configuration as `phpunit.xml.9.5.dist`
- declared all data providers as static,

Following changes have been reverted from the PR:
- modified GitHub actions to run proper PHPUnit configuration depending on the PHP version
- added PHPUnit version `^10.0` in Composer,
- added PHPUnit 10 cache dir to `.gitignore`,


nikic:
> migrated PHPUnit configuration to v10 and left old configuration as phpunit.xml.9.5.dist

Is this necessary? Doesn't PHPUnit automatically upgrade the old config?

jorgsowa:
> Is this necessary? Doesn't PHPUnit automatically migrate the old config?

XML of PHPUnit v10 is not compatible with XML of PHPUnit v9. Joining it with strictness of warnings it's not possible to use the same PHPUnit configuration file for both versions of 9 and 10.

Example of failing test with PHPUnit 10 but with elements of XML from PHPUnit 9: https://github.com/nikic/PHP-Parser/actions/runs/9086446408/job/24979454495?pr=998

The idea I came up with is to have simply two configuration files for each PHPUnit version. The other solution would be to upgrade completely to PHPUnit 10, but it would require bumping min. PHP version to 8.1 which is not desired for now.

nikic:
I think overall I'm not clear on what the benefit of using PHPUnit 10 on some versions and PHPUnit 9 on some is. This just seems to increase complexity, as we wouldn't be able to use any new functionality from PHPUnit 10 anyway. If it "just worked", that would be one thing, but apparently it doesn't...

(I think merging just the data provide changes would make sense though, in the interest of forward-compatibility.)

jorgsowa:
That makes sense. I left the code preparing codebase for the PHPUnit upgrade, but it needs to wait for a bump of a minimum PHP version to 8.1.

test/PhpParser/Parser/Php7Test.php
```diff
@@ -3,9 +3,9 @@
 namespace PhpParser\Parser;
 
 use PhpParser\Lexer;
-use PhpParser\ParserTest;
+use PhpParser\PhpVersionAbstract;
 
-class Php7Test extends ParserTest
+class Php7Test extends PhpVersionAbstract
```
nikic, line 8: `PhpVersionAbstract` here is really confusing, because there is also a `PhpVersion` class and this test has no relation to it. I think the name here should be `ParserTestAbstract`.

jorgsowa: I changed the name

nikic (approved):

Issue #1005, opened 2024-06-12
Spelling issue: dereferenceable vs dereferencable


rainx:
## What's happening

I have a question regarding another word: dereferenceable.

As a non-native English speaker, I'm not completely sure about the correct usage.

**dereferenceable** vs **dereferencable**:

When I searched for the word `dereferencable` on Wikipedia, it indicated that it is a misspelling of `dereferenceable` ([ref](https://en.wiktionary.org/wiki/dereferencable)).

However, the codebase contains both `dereferencable` and `dereferenceable`, with `dereferencable` being used more frequently.

## Actions

1. Standardize the codebase to use one term consistently.
2. Decide between using `dereferencable` or `dereferenceable` based on correctness and consistent usage.

### Recommendation

Considering the reference and the correct spelling, I recommend standardizing the term `dereferenceable` throughout the codebase.

## Related

https://github.com/nikic/PHP-Parser/pull/1004

nikic:
Fixed by #1006, thanks!

ngrams result for reference: https://books.google.com/ngrams/graph?content=dereferencable%2Cdereferenceable&year_start=1800&year_end=2019&corpus=en-2019&smoothing=3

PR #1006, opened 2024-06-12
Fix a typo: Dereferencable to Dereferenceable

rainx:
To solve the issue: https://github.com/nikic/PHP-Parser/issues/1005


rainx:
This update also references the [zend_language_parser.y](https://github.com/php/php-src/blob/14bee75960f16cc528fd1144a36a4e93628ef443/Zend/zend_language_parser.y) file from the php-src repository.

PR #1022, opened 2024-09-04
Weird token issue reproduction

ondrejmirtes:
Hi,
I'm getting a weird character at the end of tokens array with PHP-Parser 5.x.

This is how it looks like in Xdebug:

Screenshot 2024-09-04 at 23 34 10

When I'm trying to dump it:

```php
		var_dump(ord($lastToken->text));
		var_dump(bin2hex($lastToken->text));
		var_dump(strlen($lastToken->text));
```

The output looks like this:

```
int(0)
string(2) "00"
int(1)
```

Which is the same as empty string (https://3v4l.org/60Utu) except for the length.

I'm attaching a test case that shows this behaviour.

When I run the equivalent test case on PHP-Parser 4.x, the printed character is LF (newline) which is expected. But not on master (5.x).

Any help is appreciated. Thank you!

### DIFF
--- a/test/PhpParser/WeirdTokenIssueTest.php
+++ b/test/PhpParser/WeirdTokenIssueTest.php
@@ -1,0 +1,22 @@
+<?php declare(strict_types=1);
+
+namespace PhpParser;
+
+use PHPUnit\Framework\TestCase;
+use function file_get_contents;
+
+class WeirdTokenIssueTest extends TestCase
+{
+	public function testIssue(): void
+	{
+		$factory = new ParserFactory();
+		$parser = $factory->createForNewestSupportedVersion();
+		$parser->parse(file_get_contents(__DIR__ . '/ignore-line-1.php'));
+
+		$tokens = $parser->getTokens();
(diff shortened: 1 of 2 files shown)


nikic:
See https://github.com/nikic/PHP-Parser/blob/master/UPGRADE-5.0.md#changes-to-token-representation:

> The token array is now an array of Tokens, rather than an array of arrays and strings. Additionally, the token array is now terminated by a sentinel token with ID 0.

The weird character is a null byte `\0`.

Does the sentinel token cause issues for phpstan?

nikic:
> Should I just always unset the last token to get rid of this problem? Thanks.

The parser and pretty printer expect that token to be there -- but when converting to string, you should indeed ignore the last token. Basically the -1 here: https://github.com/nikic/PHP-Parser/blob/54139ca49b3b45fcf09257727611caa52cd0c00b/lib/PhpParser/PrettyPrinterAbstract.php#L547

ondrejmirtes:
Yeah, thanks, this is sufficient :)

PR #1025, opened 2024-09-17
Update FAQ.markdown to use addVisitor instead of constructor.

chx:


nikic:
Why?

chx:
```
    public function __construct() {
        // for BC
    }
```

chx:
[](url)

Issue #1028, opened 2024-09-29
php-parser 4.19.3 cause error Type of PhpParser\Parser\Php8::$tokenToSymbolMapSize must not be defined (as in class PhpParser\ParserAbstract)


samsonasik:
Just tried in rector-src

```
PHP Fatal error:  Type of PhpParser\Parser\Php8::$tokenToSymbolMapSize must not be defined (as in class PhpParser\ParserAbstract) in /home/runner/work/rector-src/rector-src/vendor/nikic/php-parser/lib/PhpParser/Parser/Php8.php on line 19
```

see https://github.com/rectorphp/rector-src/actions/runs/11093478852/job/30819810695?pr=6336#step:8:10

PR #1029, opened 2024-09-29
Fix protected property type error from ParserAbstract on Php8.php on 4.19.3

samsonasik:
version 4.19.3 cause error:

```
PHP Fatal error:  Type of PhpParser\Parser\Php8::$tokenToSymbolMapSize must not be defined (as in class PhpParser\ParserAbstract) in /home/runner/work/rector-src/rector-src/vendor/nikic/php-parser/lib/PhpParser/Parser/Php8.php on line 19
```

This PR fix it.

Fixes https://github.com/nikic/PHP-Parser/issues/1028

### DIFF
--- a/lib/PhpParser/Parser/Php8.php
+++ b/lib/PhpParser/Parser/Php8.php
@@ -163,19 +163,19 @@ class Php8 extends \PhpParser\ParserAbstract
     public const T_NAME_RELATIVE = 398;
     public const T_ATTRIBUTE = 399;
 
-    protected int $tokenToSymbolMapSize = 400;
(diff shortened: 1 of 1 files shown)


samsonasik:
Ready to merge 👍

nikic:
Ooops, I made a mistake when backporting a change from master. This file shouldn't exist at all.

nikic:
I removed the file and released 4.19.4. Does everything work now?

samsonasik:
testing it at https://github.com/rectorphp/rector-src/pull/6338

samsonasik:
it seems working ok now, thank you 👍

Issue #1033, opened 2024-10-06
Add `Node->getSubNodes()`


staabm:
the current `Node->getSubNodeNames()` api leads to consuming code with dynamic property fetches like:

```
			foreach ($node->getSubNodeNames() as $subNodeName) {
				if ($node instanceof Node\Expr\Closure && $subNodeName !== 'uses') {
					continue;
				}
				$subNode = $node->{$subNodeName};
				$variableNames = array_merge($variableNames, $this->getUsedVariables($scope, $subNode));
			}
```

In PHPStan-src alone this pattern repeats 8 times. 

Its also visibile in the php-parser codebase 2 times:
https://github.com/nikic/PHP-Parser/blob/3abf7425cd284141dc5d8d14a9ee444de3345d1a/lib/PhpParser/PrettyPrinterAbstract.php#L643-L651
https://github.com/nikic/PHP-Parser/blob/3abf7425cd284141dc5d8d14a9ee444de3345d1a/lib/PhpParser/NodeTraverser.php#L93-L98

this dynamic property fetches lead to badly static analysable code

> as soon as a class contains a single dynamic property lookup, phpstan is no longer able to tell you whether one of the declared properties is unused (analog for methods).

what do you think about adding a new `Node->getSubNodes()` method, so the dynamic property lookup is only available in one place and caller code would be more static analysis friendly?

staabm:
I have something prepared which can make the code static analysis friendly and does not require a new method. stay tuned.

PR #1035, opened 2024-10-07
SubNodeNames are known at static analysis time

staabm:
this PR turns the SubNodeNames into constant arrays and corresponding phpdocs, which makes them available to [PHPStan at static analysis time](https://phpstan.org/r/67c66b26-af66-4ec0-ae3c-cb0eac3e9b4d).

this information is useful to better cover codebases which utilize PHP-Parser 

closes https://github.com/nikic/PHP-Parser/issues/1033


nikic:
I can kind of see how this could be useful in theory, but it's not really clear to me where we'd see a benefit in practice. Can you share an example where e.g. this gets rid of a baseline suppression in phpstan-src or similar?

staabm:
see e.g. https://phpstan.org/r/9041358f-043d-4171-9780-f9d9b4b3c534

before this PR here, the unused property cannot be detected.

after this PR we get
- as initially suggsted, the unused property now gets detected
- the error on line 26 is a unrelated PHPStan bug

```
➜  phpstan-src git:(2.0.x) ✗ php bin/phpstan analyze --debug test.php
Note: Using configuration file /Users/staabm/workspace/phpstan-src/phpstan.neon.dist.
/Users/staabm/workspace/phpstan-src/test.php
 ------ ---------------------------------------------------------------------------------------------------- 
  Line   test.php                                                                                            
 ------ ---------------------------------------------------------------------------------------------------- 
  11     Property Processor::$unusedProperty is never read, only written.                                    
         🪪  property.onlyWritten                                                                            
         💡 See: https://phpstan.org/developing-extensions/always-read-written-properties                    
  26     Access to an undefined property PhpParser\Node\Expr\FuncCall|PhpParser\Node\Expr\MethodCall::$var.  
         🪪  property.notFound                                                                               
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property                
 ------ ---------------------------------------------------------------------------------------------------- 

```

staabm:
hmm.. I tried this PR on the whole phpstan-src codebase and it somehow has strange side-effects. I have another look

lib/PhpParser/Internal/PrintableNewAnonClassNode.php
```diff
@@ -29,6 +29,8 @@ class PrintableNewAnonClassNode extends Expr {
     /** @var Node\Stmt[] Statements */
     public array $stmts;
 
+    private const SUBNODE_NAMES = ['attrGroups', 'flags', 'args', 'extends', 'implements', 'stmts'];
```
mvorisek, line 32: It would be great to assert the lists with a test, so the lists are never outdated/incomplete.

PR #1038, opened 2024-10-23
 Allow TRUE instead of \TRUE

chx:
I am working on a Rector rule which converts a very large chunk of the remaining procedural code in Drupal to OOP. (Run!)

One problem we found the converted code contains \TRUE instead of TRUE. While of course \TRUE is totally correct, people wanted TRUE and who am I to disagree? I added this to a visitor:

```
                    if (isset($node->name) && $node->name instanceof FullyQualified) {
                        $name = new Node\Name($node->name);
                        if ($name->isUnqualified()) {
                            $node->name = $name;
                            return $node;
                        }
                    }
```

that's quite the hassle though. So I am proposing a property on FullyQualified to start the discussion and if you find it acceptable then I'd add an option to NameResolver which sets this to FALSE. Thanks for your consideration -- and the amazing library.

### DIFF
--- a/lib/PhpParser/Node/Name/FullyQualified.php
+++ b/lib/PhpParser/Node/Name/FullyQualified.php
@@ -3,6 +3,9 @@
 namespace PhpParser\Node\Name;
 
 class FullyQualified extends \PhpParser\Node\Name {
+
+    public bool $anchorRoot = TRUE;
+    
     /**
      * Checks whether the name is unqualified. (E.g. Name)
      *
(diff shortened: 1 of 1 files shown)


nikic:
Can you please clarify where the `\TRUE` is coming from?

Generally, the correct way to handle this is by not creating a FullQualified name in the first place, rather than messing with how it is printed.

chx:
Well, rector gives me fully qualified nodes. I am just writing a rector rule using what's given to me. And it's actually quite useful for everything except for things from the root. I guess then the solution would be to find out where fullyqualified comes from  which is hopeless. OK.

nikic:
> Well, rector gives me fully qualified nodes. I am just writing a rector rule using what's given to me. And it's actually quite useful for everything except for things from the root. I guess then the solution would be to find out where fullyqualified comes from which is hopeless. OK.

If getting unqualified names is too much of a hassle, the alternative would be to extend the used PrettyPrinter and override pName_FullyQualified there with your special case.

PR #1040, opened 2024-11-18
update target.php to new PHP-Parser version

jorgsowa:
Updates fuzzer target.php to new version of PHP-Parser. The method getTokens() has been removed.

The PHPStan issue is not related to the PR.


tools/fuzzing/target.php
```diff
@@ -104,7 +104,7 @@ public function leaveNode(PhpParser\Node $node) {
     $stmts = $parser->parse($input);
     $printed = $prettyPrinter->prettyPrintFile($stmts);
 
-    $visitor->setTokens($lexer->getTokens());
+    $visitor->setTokens($lexer->tokenize($input));
```
nikic, line 107: This should be `$parser->getTokens()`, no need to tokenize a second time.

jorgsowa: Replaced

Issue #1050, opened 2024-12-11
Property with multiple props and property hooks should be a parse error


ondrejmirtes:
This piece of code:

```php
class Foo
{
    
    public $foo, $bar { get {} }
    
}
```

Is rejected by PHP with error message: `Parse error: syntax error, unexpected token “{”, expecting “,” or “;” in /in/RndUG on line 6`

But PHP-Parser does not report syntax error and represents this with an AST.

Multi-props with property hooks are not supported in PHP. I'm asking about this to find out if this should be a parse error in PHP-Parser, or if this is going to be represented with an AST and PHPStan should come with a custom rule to detect it.

Thanks.

nikic:
I'm fine with having this as a parse error.

ondrejmirtes:
Thanks. Unfortunately I don't know how to implement this one, grammar is still out of my expertise.

nikic:
This one wouldn't be checked directly in the grammar, it needs a manual check like the one here: https://github.com/nikic/PHP-Parser/blob/4f9dc8b0f50d7bbc6b3197ec5cb2ad928fc10d40/lib/PhpParser/ParserAbstract.php#L1161-L1167

ondrejmirtes:
Alright, tried it: https://github.com/nikic/PHP-Parser/pull/1052

Had to modify the grammar anyway 😅

nikic:
Fixed by https://github.com/nikic/PHP-Parser/commit/d20a197ca73fc07d1aa980509b49f4ce268a22a4.

PR #1052, opened 2024-12-11
Reject multi hooked property

ondrejmirtes:
Closes https://github.com/nikic/PHP-Parser/issues/1050

### DIFF
--- a/lib/PhpParser/ParserAbstract.php
+++ b/lib/PhpParser/ParserAbstract.php
@@ -1159,7 +1159,7 @@ abstract class ParserAbstract implements Parser {
     }
 
     /** @param PropertyHook[] $hooks */
-    protected function checkPropertyHookList(array $hooks, int $hookPos): void {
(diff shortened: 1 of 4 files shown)


lib/PhpParser/ParserAbstract.php
```diff
@@ -1158,8 +1158,15 @@ protected function checkUseUse(UseItem $node, int $namePos): void {
         }
     }
 
+    protected function checkPropertyHooksForMultiProperty(Property $property): void {
+        if (count($property->props) > 1) {
+            $this->emitError(new Error(
+                'Multiple properties cannot share the same hooks', $property->getAttributes()));
```
nikic, line 1164: This is marking the entire property as the error location. This should probably use the opening brace of the hooks instead, like the one below. (Or alternatively the name of the second property, not sure which is better.)

ondrejmirtes:
Thank you!

Issue #1055, opened 2024-12-30
Inefficient NodeConnectingVisitor


staabm:
when using [`NodeConnectingVisitor`](https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/NodeVisitor/NodeConnectingVisitor.php) most AST nodes get additional attributes, which turns GC inefficient.

PHPStan enabled a optimization in https://phpstan.org/blog/preprocessing-ast-for-custom-rules to get rid of `NodeConnectingVisitor` and I also got [reports in phpstan-dba](https://github.com/staabm/phpstan-dba/issues/667) which shows concrete profile numbers about this very same problem (I think).

is there anything which can be improved in NodeConnectingVisitor to better support GC?

nikic:
NodeConnectingVisitor adds a parent reference, which introduces a cycle. This means that the AST can now only be collected by cycle GC. As adding the parent reference is the point of the visitor, there's nothing that can be done about it. This is also why the AST does not provide parent references by default and requires the opt-in via the visitor.

The only improvement I can think of here would be to allow explicitly destroying AST nodes, which would clear out the parent reference and break the cycle. This would have to be done explicitly by the user though.

staabm:
Maybe this "side-effects" should be mentioned in the class comment, so people get aware of it?

I am fine ic there is no room for improvement.

In my case I copied the visitor and reduced it to the props I really need to make the problem less impactful

nikic:
Thinking about this again, a possible improvement would be to store the parent reference as a WeakReference. That will make usage a bit more awkward, but avoid the cycle. I can't change the existing visitor, but maybe could add an option to use a weak reference?

staabm:
why does only the parent reference create a cycle? wouldn't previous/next refernce also create a cycle of nodes (which all have the same parent)?

nikic:
Yes, previous/next also create a cycle (if both a present, it's not a cycle if there's only one). I'd switch all three to use WeakReference if going down this path.

PR #1057, opened 2025-01-01
Support WeakReferences in NodeConnectingVisitor

staabm:
closes https://github.com/nikic/PHP-Parser/issues/1055

we observed perf improvements on a similar visitor in https://github.com/staabm/phpstan-dba/issues/667#issuecomment-2565559874

### DIFF
--- a/doc/component/FAQ.markdown
+++ b/doc/component/FAQ.markdown
@@ -51,3 +51,17 @@ obtained through `$node->getAttribute('next')`.
 
 `ParentConnectingVisitor` and `NodeConnectingVisitor` should not be used at the same time. The latter
 includes the functionality of the former.
+
+
+How can I limit the impact of cyclic references in the AST?
+-----
+
+NodeConnectingVisitor adds a parent reference, which introduces a cycle. This means that the AST can now only be collected by cycle garbage collector.
+This in turn can lead to performance and/or memory issues. 
+
+To break the cyclic references between AST nodes `NodeConnectingVisitor` supports a boolean `$weakReferences` constructor parameter.
+When set to `true`, all attributes added by `NodeConnectingVisitor` will be wrapped into a `WeakReference` object.
+
+After enabling this parameter, the parent node can be obtained through `$node->getAttribute('weak_parent')`,
+the previous node can be obtained through `$node->getAttribute('weak_previous')`, and the next node can be
+obtained through `$node->getAttribute('weak_next')`.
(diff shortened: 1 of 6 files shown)


staabm:
I think in CloningVisitor we don't need a weak-ref because we clone the elements, right?

lib/PhpParser/NodeVisitor/NodeConnectingVisitor.php
```diff
@@ -25,19 +25,34 @@ final class NodeConnectingVisitor extends NodeVisitorAbstract {
...
     }
 
     public function enterNode(Node $node) {
         if (!empty($this->stack)) {
-            $node->setAttribute('parent', $this->stack[count($this->stack) - 1]);
+            $parent = $this->stack[count($this->stack) - 1];
+            if ($this->weakReferences) {
+                $parent = \WeakReference::create($parent);
+            }
+            $node->setAttribute('parent', $parent);
```
nikic, line 45: One thing I'm wondering about here is whether we should reuse the same attribute names, or use a different one like `'weak_parent'`, to avoid having two possible types for these attributes.

staabm: We could use separate attribute names, but then I would go with a separate  WeakNodeConnectingVisitor class and leave  NodeConnectingVisitor untouched ..?

staabm: adjusted

nikic: Given that the implementation themselves are pretty trivial, I'd be fine with adding separate classes as well.

staabm: I am fine with the PR as is, in case you are also ok with it

staabm:
Friendly ping

doc/component/FAQ.markdown
```diff
@@ -51,3 +51,17 @@ obtained through `$node->getAttribute('next')`.
 
 `ParentConnectingVisitor` and `NodeConnectingVisitor` should not be used at the same time. The latter
 includes the functionality of the former.
+
+
+How can I limit the impact of cyclic references in the AST?
+-----
+
+NodeConnectingVisitor adds a parent reference, which introduces a cycle. This means that the AST can now only be collected by cycle garbage collector.
```
nikic, line 59:
```suggestion
NodeConnectingVisitor adds a parent reference, which introduces a cycle. This means that the AST can now only be collected by the cycle garbage collector.
```

PR #993, opened 2024-04-12
Declare more precise phpdoc types

staabm:
goal: ease working in projects which are static analysis covered.

we are doing the same thing in PHPStan with https://github.com/phpstan/phpstan-src/pull/3013#pullrequestreview-1998289070

PR #1058, opened 2025-01-01
More precise phpdocs

staabm:
also fixes a few PHPStan level 7 errors

### DIFF
--- a/lib/PhpParser/Internal/TokenPolyfill.php
+++ b/lib/PhpParser/Internal/TokenPolyfill.php
@@ -20,9 +20,9 @@ class TokenPolyfill {
     public int $id;
     /** @var string The textual content of the token. */
     public string $text;
-    /** @var int The 1-based starting line of the token (or -1 if unknown). */
+    /** @var -1|positive-int The 1-based starting line of the token (or -1 if unknown). */
     public int $line;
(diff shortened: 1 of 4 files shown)


nikic (commented):
Per the comments on https://github.com/nikic/PHP-Parser/pull/993 it was my understanding that we shouldn't change argument types to avoid BC breaks. Not sure to what degree this matters here.

staabm:
you are right.. param types should only be done on new major versions. dropped the param type changes

lib/PhpParser/Internal/TokenPolyfill.php
```diff
@@ -20,9 +20,9 @@ class TokenPolyfill {
     public int $id;
     /** @var string The textual content of the token. */
     public string $text;
-    /** @var int The 1-based starting line of the token (or -1 if unknown). */
+    /** @var -1|positive-int The 1-based starting line of the token (or -1 if unknown). */
     public int $line;
-    /** @var int The 0-based starting position of the token (or -1 if unknown). */
+    /** @var int<-1, max> The 0-based starting position of the token (or -1 if unknown). */
```
nikic, line 25: Do these types match whatever phpstan specifies for the PhpToken class? As we may either use this class or PhpToken. Not sure which one phpstan would actually pick for analysis...

PR #1061, opened 2025-01-17
Remove useless foreach key and cast string, and fix invalid @var definition on test

samsonasik:


samsonasik:
Ready to merge 👍

test/PhpParser/Node/Scalar/DNumberTest.php
```diff
@@ -18,7 +18,7 @@ public function testRawValue(): void {
         $lLumber = $echo->exprs[0];
```
nikic, line 18:
```suggestion
        $dnumber = $echo->exprs[0];
```
We should change the variable instead.

test/PhpParser/Node/Scalar/NumberTest.php
```diff
@@ -17,7 +17,7 @@ public function testRawValue(): void {
         $lLumber = $echo->exprs[0];
```
nikic, line 17:
```suggestion
        $lnumber = $echo->exprs[0];
```

samsonasik:
updated the variable names 👍

PR #1067, opened 2025-01-28
Fix name context on no namespace

samsonasik:
Ref https://getrector.com/ast/d50f4d493f7061f3c419397a744a49e92409c3ce?nodeId=1

which `\true` should be `true` on no namespace.

### DIFF
--- a/test/PhpParser/NodeVisitor/NameResolverTest.php
+++ b/test/PhpParser/NodeVisitor/NameResolverTest.php
@@ -117,8 +117,8 @@ namespace {
     new \Hallo\Bar();
     new \Bar();
     new \Bar();
-    \bar();
-    \hi();
+    bar();
+    hi();
     \Hallo\bar();
     \foo\bar();
     \bar();
@@ -197,7 +197,7 @@ class A extends B implements C, D {
 
     #[X]
     const C = 1;
-    
+
     public const X A = X::Bar;
     public const X\Foo B = X\Foo::Bar;
     public const \X\Foo C = \X\Foo::Bar;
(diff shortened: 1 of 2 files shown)


samsonasik:
Ready to merge 👍

nikic (commented):
I don't think this change is correct. Outside a namespace we should still fully resolve to full qualified names. Unqualified names should only be left if resolution is impossible (due to global fallback) or because of special class names.

samsonasik:
the verification of fully qualified already checked before check, so I think that's safe

test/PhpParser/NodeVisitor/NameResolverTest.php
```diff
@@ -117,8 +117,8 @@ public function testResolveNames(): void {
     new \Hallo\Bar();
     new \Bar();
     new \Bar();
-    \bar();
-    \hi();
+    bar();
```
nikic, line 120: This is what I mean. This should resolve to a fully qualified name. The fact that it is equivalent to the unqualified name doesn't matter, we should always resolve fully if it is possible.

samsonasik: if i understand correctly, the test is removing namespace, which no longer has namespace, means it doesn't have `\\` prefix, next line to it, it have `\bar` which not changed as already fully qualified on the first place, which expexted.

samsonasik: see line 124 which kept to have `\\` prefix as it previously already fully qualified.

PR #1075, opened 2025-03-11
Add option to disable recursive refs in Name Resolution

dhaeckel:


nikic:
Hm, I wonder if we should just always do the clone if replaceNodes is disabled, rather than introduce an additional option.

dhaeckel:
Happy to change the PR accordingly, although @theofidry raised the concern of higher memory consumption in this case. This would impact all apps that use replaceNodes = false. I can't really tell, if this is a problematic impact or negligible. I'll leave it up to you, which way you prefer.

dhaeckel:
Hi @nikic,

do you have any opinion on how to go forward with this? I'm happy with any of the two solutions.

PR #1080, opened 2025-04-10
Declare CallLike $args as list

staabm:
having a more precise type helps to prevent false positives when implementing e.g. PHPStan rules

```
private function findInputTypeIdsInNew(New_ $newExpr, ClassReflection $class, Scope $scope): array
    {
        if (\count($newExpr->args) === 0) {
            // more logic
            return [];
        }

        $firstArg = $newExpr->args[0]; // Offset 0 might not exist on non-empty-array<PhpParser\Node\Arg|PhpParser\Node\VariadicPlaceholder>.
        if (! $firstArg instanceof Arg) {
            return [];
        }
```

happens when PHPStan is configured with [reportPossiblyNonexistentGeneralArrayOffset](https://phpstan.org/config-reference#reportpossiblynonexistentgeneralarrayoffset)

### DIFF
--- a/lib/PhpParser/Node/Expr/FuncCall.php
+++ b/lib/PhpParser/Node/Expr/FuncCall.php
@@ -8,14 +8,14 @@ use PhpParser\Node\Expr;
 class FuncCall extends CallLike {
     /** @var Node\Name|Expr Function name */
     public Node $name;
-    /** @var array<Node\Arg|Node\VariadicPlaceholder> Arguments */
+    /** @var list<Node\Arg|Node\VariadicPlaceholder> Arguments */
     public array $args;
 
     /**
      * Constructs a function call node.
      *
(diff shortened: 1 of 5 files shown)


nikic:
This looks fine, but will this cause BC issues? I think people were concerned about that in the past.

staabm:
yes, I had similar feelings when submitting the PR. When in doubt this PR has to wait for the next major version

staabm:
@nikic today I stumbled over this again while working on phpstan-src.

do you think it makes sense merging?

staabm:
alternatively (and more safe) we could re-activate https://github.com/phpstan/phpstan-php-parser to override the signatures in phpstan context

staabm:
do it here instead: https://github.com/phpstan/phpstan-src/pull/4503

PR #1082, opened 2025-04-29
Add support for attributes on constants

DanielEScherzer:
Just merged in php/php-src@3f03f7ed3d988567b5a59ae542579fd91cdfde42, updating this parser is needed to be able to use attributes on constants in the stub files.


lib/PhpParser/Node/Stmt/Const_.php
```diff
@@ -7,20 +7,28 @@
...
      * Constructs a const list node.
      *
      * @param Node\Const_[] $consts Constant declarations
+     * @param list<Node\AttributeGroup> $attrGroups PHP attribute groups
      * @param array<string, mixed> $attributes Additional attributes
      */
-    public function __construct(array $consts, array $attributes = []) {
+    public function __construct(
+        array $consts,
+        array $attrGroups = [],
```
nikic, line 22: Please move $attrGroups after $attributes to maintain backwards compatibility.

nikic (commented):
Please also add pretty printer support and a pretty printer test.

This will also need an extra resolveAttrGroups call in NameResolver.

(See https://github.com/nikic/PHP-Parser/blob/master/CONTRIBUTING.md#adding-support-for-new-php-syntax for the full checklist.)

lib/PhpParser/Node/Stmt/Const_.php
DanielEScherzer: Except the callback in resolveNodes() always passes the attributes at the end - should I add special handling there?

nikic: Use `new Stmt\Const_($3, attributes(), $1)` instead.

DanielEScherzer: Done

DanielEScherzer:
> Please also add pretty printer support and a pretty printer test.
> 
> This will also need an extra resolveAttrGroups call in NameResolver.
> 
> (See https://github.com/nikic/PHP-Parser/blob/master/CONTRIBUTING.md#adding-support-for-new-php-syntax for the full checklist.)

Done, thanks for the pointer

nikic (approved):

grammar/php.y
```diff
@@ -257,7 +257,10 @@ top_statement:
     | T_USE use_declarations semi                           { $$ = Stmt\Use_[$2, Stmt\Use_::TYPE_NORMAL]; }
     | T_USE use_type use_declarations semi                  { $$ = Stmt\Use_[$3, $2]; }
     | group_use_declaration
-    | T_CONST constant_declaration_list semi                { $$ = Stmt\Const_[$2]; }
+    | T_CONST constant_declaration_list semi                { $$ = Stmt\Const_[$2, attributes(), []]; }
```
nikic, line 260:
```suggestion
    | T_CONST constant_declaration_list semi                { $$ = new Stmt\Const_($2, attributes(), []); }
```

grammar/php.y
```diff
@@ -257,7 +257,10 @@ top_statement:
...
+    | attributes T_CONST constant_declaration_list semi
+          { $$ = Stmt\Const_[$3, attributes(), $1];
```
nikic, line 262:
```suggestion
          { $$ = new Stmt\Const_($3, attributes(), $1);
```

nikic (commented):
This should fix the phpstan errors I believe.

lib/PhpParser/PrettyPrinterAbstract.php
```diff
@@ -1634,6 +1634,7 @@ protected function initializeEmptyListInsertionMap(): void {
             Stmt\Trait_::class . '->attrGroups' => [null, '', "\n"],
             Expr\ArrowFunction::class . '->attrGroups' => [null, '', ' '],
             Expr\Closure::class . '->attrGroups' => [null, '', ' '],
+            Expr\Const_::class . '->attrGroups' => [null, '', "\n"],
```
nikic, line 1637:
```suggestion
            Stmt\Const_::class . '->attrGroups' => [null, '', "\n"],
```

nikic:
Also need to regenerate the parsers.

PR #1092, opened 2025-07-02
Support every argument syntax for `clone()`

TimWolla:
see php/php-src#18938


grammar/php.y
```diff
@@ -785,6 +785,22 @@ argument_list:
...
+clone_argument_list:
+      '(' ')'                                              { $$ = array(); }
+    | '(' non_empty_clone_argument_list optional_comma ')' { $$ = $2; }
+    | '(' expr ',' ')'                                     { init(Node\Arg[$2, false, false]); }
+    | '(' variadic_placeholder ')'                         { init($2); }
+;
+
+non_empty_clone_argument_list:
+		expr ',' argument
+			{ init(new Node\Arg($1, false, false, stackAttributes(#1)), $3); }
```
nikic, line 797: Can you add a `!!positions` test to check the two cases using stackAttributes?

PR #1098, opened 2025-08-06
Add cast kinds for bool, int, string

mttsch:


test/code/prettyPrinter/expr/cast.test
```diff
@@ -0,0 +1,17 @@
...
+
+$int = (int) $int;
+$integer = (integer) $integer;
+$bool = (bool) $bool;
+$boolean = (boolean) $boolean;
+$string = (string) $string;
+$binary = (binary) $binary;
+-----
+$int = (int) $int;
+$integer = (int) $integer;
```
mttsch, line 13: I kept the previous behavior that converts the aliased casts to their un-alised casts for backward compatibility while writing. This is in contrast to double-like casts, where `pExpr_Cast_Double()` considers the original kind.

nikic:
This is to detect deprecated cast kinds?

mttsch:
@nikic Yes, exactly.

staabm:
@mttsch could you check whether this PR broke the PHP8.5 unit tests (see latest master commit which added PHP8.5 coverage after this PR)?

mttsch:
@staabm The issue does not exist for the tagged alpha4 version:

```
root@f056498d2af3:/var/www/html# php -v
PHP 8.5.0alpha4 (cli) (built: Aug 13 2025 03:24:10) (NTS)
Copyright (c) The PHP Group
Built by https://github.com/docker-library/php
Zend Engine v4.5.0-dev, Copyright (c) Zend Technologies
    with Zend OPcache v8.5.0alpha4, Copyright (c), by Zend Technologies
root@f056498d2af3:/var/www/html# ./vendor/bin/phpunit 
PHPUnit 9.6.20 by Sebastian Bergmann and contributors.

.............................................................   61 / 1874 (  3%)
[…]
............................................                  1874 / 1874 (100%)

Time: 00:00.530, Memory: 44.00 MB

OK (1874 tests, 2747 assertions)
```

I am not sure if I want to/can spend the time to set up a PHP setup/container with the latest commit (or same commit as in CI) from the php-src at the moment.

PS: … I will try setting it up this evening if time permits.

nikic:
That was due to a php-src bug that I already fixed in https://github.com/php/php-src/pull/19438.

staabm:
awesome thanks for the feedback

PR #1100, opened 2025-08-13
Drop branch-alias on 4.x

nicolas-grekas:
A similar fix is needed on master

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,11 +20,6 @@
         "phpunit/phpunit": "^7.0 || ^8.0 || ^9.0",
         "ircmaxell/php-yacc": "^0.0.7"
     },
-    "extra": {
-        "branch-alias": {
-            "dev-master": "4.9-dev"
-        }
-    },
     "autoload": {
         "psr-4": {
             "PhpParser\\": "lib/PhpParser"


nikic (commented):
I think this branch alias should just be dropped. Especially on the 4.x branch having a dev-master alias doesn't really make sense?

nicolas-grekas:
of course! PR updated.

pk-fr:
this change breaks my php obfuscator :

I need a way to know which PHP-Parser version is installed to make sure that the version is ok with the obfuscator...
I was using branch-alias to determine the version...
Please offer a way to do that: for exemple a version.php file 

thanx in advance!

nicolas-grekas:
You should use the `Composer\InstalledVersions` class for that. It's in your vendor/composer directory.

pk-fr:
I do not use composer!
I use git clone

nikic:
Parsing the changelog may be an option.

pk-fr:
> Parsing the changelog may be an option.

It is the work-around that I will have to do, if PHP-Parser do not provide easy and standard way for getting it's version...
As you may imagine, I would prefer just having an include file that will have a variable or a function getting the version!

PR #1104, opened 2025-08-19
Deprecate Double cast

mttsch:
With PHP's `(double)` cast being deprecated and already having only been an alias for the `(float)` cast, introduce `Float_` as a replacement for the `Double` cast.

### DIFF
--- a/grammar/php.y
+++ b/grammar/php.y
@@ -1110,7 +1110,7 @@ expr:
     | T_DOUBLE_CAST expr
           { $attrs = attributes();
             $attrs['kind'] = $this->getFloatCastKind($1);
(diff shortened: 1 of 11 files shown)


mttsch:
Primarily, this PR is meant as a discussion opener whether such a deprecation is even wanted or deemed too disruptive for package consumers.

nikic:
We can only rename nodes at major version boundaries. I'd suggest filing an issue and we can add it to the 6.0 milestone.

mttsch:
@nikic Done #1105 1105

vinay070403:
👍 Makes sense to align with PHP’s deprecation of `(double)`.  
Happy to help test this once it lands in the 6.0 branch to check compatibility with downstream packages.

PR #1107, opened 2025-09-05
Enhance ConstExprEvaluator by evaluating new type of expressions or enhancing some (hopefully)

odain-cbd:
For iTop project we had to enhance ConstExprEvaluator to parse/evaluate few of our PHP files. Below are listed the added/enhanced Expression usecases.
(https://github.com/Combodo/iTop/tree/feature/4789-nikicfork)

Added:
	- Variable
	- Isset
	- ClassConstFetch
	- Cast
	- StaticPropertyFetch
	- FuncCall
	- StaticCall
	- NullsafePropertyFetch
	- PropertyFetch
	- NullsafeMethodCall
	- MethodCall

Enhanced:
	- ConstFetch
	- Coalesce


nikic (changes requested):
The purpose of ConstExprEvaluator is to evaluate "constant expressions", such as the right hand side of `const X = ...`. These constant expressions only support a limited set of operations.

What you seem to be implementing here seems to be a lot broader: It's trying to evaluate *all* expressions, and accessing global state (like global variables) for that purpose. While this is may be useful for some use cases, it does not fit the purpose of ConstExprEvaluator and this behavior would be highly undesirable in most circumstances (static analyzers do not typically run in the same execution context as the code they're analyzing, so this just ends up accessing global state from the analyzer).

odain-cbd:
Thanks for the feedback.

Keeping the purpose of ConstExprEvaluator, I left only below enhancements in this class:
ConstFetch/ ClassConstFetch (added)/ Cast (added)

All non-constant evaluations have been moved in ExprEvaluator class. with associated tests as well.

lib/PhpParser/ConstExprEvaluator.php
```diff
@@ -145,6 +143,14 @@ private function evaluate(Expr $expr) {
             return $this->evaluateConstFetch($expr);
         }
 
+	    if ($expr instanceof Expr\ClassConstFetch) {
```
staabm, line 146: mix of tabs and spaces in all changed files

odain-cbd: sorry for that. just fixed it

Molkobain:
Hello @nikic , could this be reconsidered please? Thanks :)

Issue #1119, opened 2025-10-18
Format-preserving printer loses parentheses around BooleanNot expressions


maks-rafalko:
### Environment

- **PHP-Parser Version:** 5.6.1
- **PHP Version:** 8.4.4
- Reproducer: https://github.com/nikic/PHP-Parser/pull/1120

### Description

When I wrap a `BinaryOp\BooleanAnd` node in a `BooleanNot` node, the format-preserving printer **loses the required parentheses** around the boolean AND expression. This produces semantically incorrect code.

### Expected Behavior

The format-preserving printer should add parentheses around the wrapped expression:

```php
$var = !(a() && b());
```

## Actual Behavior

The parentheses are missing, changing the code's meaning:

```php
$var = !a() && b();
```

### Mutation

```
$stmts[0]->expr->expr = new \PhpParser\Node\Expr\BooleanNot($booleanAnd);
```

Enfiladeguide:
Another instance of the same root cause, in case it helps prioritise: with php-parser 5.8.0 (still the case in 5.9.0), a `Ternary` moved unchanged into the right operand of a new `Coalesce` is printed by `printFormatPreserving()` as `$a ?? $b ? 'yes' : 'no'`, while `prettyPrint()` gives `$a ?? ($b ? 'yes' : 'no')`. The same happens with a new `Minus` over an original `($b - $c)`: `$a - $b - $c` instead of `$a - ($b - $c)`. Downstream this surfaced as a semantics-changing rewrite in Rector's TernaryToNullCoalescingRector (reported separately).

PR #1120, opened 2025-10-18
Reproducer for a bug with format-preserving printer that loses parentheses around BooleanNot expressions

maks-rafalko:
This is only a reproducer for the issue reported in #1119

### DIFF
--- a/test/code/formatPreservation/booleanReproducer.test
+++ b/test/code/formatPreservation/booleanReproducer.test
@@ -1,0 +1,12 @@
+Reproducer for a bug with boolean negations
+-----
+<?php
+
+$a = a() && b();
+-----
+$booleanAnd = $stmts[0]->expr->expr;
+$stmts[0]->expr->expr = new \PhpParser\Node\Expr\BooleanNot($booleanAnd);
+-----
+<?php
+
+$a = !(a() && b());


test/code/formatPreservation/booleanReproducer.test
```diff
@@ -0,0 +1,12 @@
...
+<?php
+
+$a = a() && b();
+-----
+$booleanAnd = $stmts[0]->expr->expr;
+$stmts[0]->expr->expr = new \PhpParser\Node\Expr\BooleanNot($booleanAnd);
+-----
+<?php
+
+$a = !(a() && b());
```
maks-rafalko: Fails with

```diff
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
 '<?php
 
-$a = !(a() && b());'
+$a = !a() && b();'
```

nikic:
PHP-Parser 5 uses a different way to handle operator precedence in the pretty printer, and it doesn't interact with formatting preservation properly :/

nikic:
This looks pretty tricky. Previously responsibility for printing the parentheses was with the parent node, now it's with the child. Which only works properly if the child is printed without formatting.

PR #1122, opened 2025-10-20
Add getVisitorsForNode() method to NodeTraverser

samsonasik:
@TomasVotruba here for your request https://github.com/rectorphp/rector-src/pull/6232#issuecomment-3423162697

This is template config for `getVisitorsForNode()`, which on rector use case, we use for caching visitor by node.

https://github.com/rectorphp/rector-src/blob/4b30bc5d10e7c79867a6a13da3a75a1b2c844675/src/PhpParser/NodeTraverser/RectorNodeTraverser.php#L65-L82

so we can just override the method there :), we currently using vendor-patch for it:

https://github.com/rectorphp/vendor-patches/blob/18e174cbea7beaccb140cfcf03cca3097e95ee92/patches/nikic-php-parser-lib-phpparser-nodetraverser-php.patch

Original PR for this usage for caching on Rector side:

- https://github.com/rectorphp/rector-src/pull/6232

### DIFF
--- a/lib/PhpParser/NodeTraverser.php
+++ b/lib/PhpParser/NodeTraverser.php
@@ -275,7 +275,7 @@ class NodeTraverser implements NodeTraverserInterface {
      *
      * @return NodeVisitor[]
      */
-    protected function getVisitorsForNode(Node $node) {
+    protected function getVisitorsForNode(Node $node): array {
         return $this->visitors;
     }
 


samsonasik:
Ready for review 👍

carlos-granados:
In the case of Rector, this was used to optimise the running time by 20-30%. Probably other tools could benefit from similar optimisations

TomasVotruba:
The 20-30 % speedup was really noticable across many mid-large size projects :+1: 
It would also open-up even more advanced speed improvements based on full node class -> visitor tree.

lib/PhpParser/NodeTraverser.php
```diff
@@ -268,6 +270,13 @@ protected function traverseArray(array $nodes): array {
         return $nodes;
     }
 
+    /**
+     * @return NodeVisitor[]
+     */
+    public function getVisitorsForNode(Node $node) {
```
nikic: Should probably the protected and have array return type?

Would also be good to add a doc comment to explain that this is an extension point, as the method doesn't make a lot of sense in itself.

nikic:
It looks like adding this method makes NodeTraverser itself about 10% slower.

samsonasik:
@nikic could you share script to verify the benchmark? Thank you.

samsonasik:
@nikic I created benchmark repo for non-cached vs cached node visitor version

https://github.com/samsonasik/php-parser-with-cache-node-visitor-benchmark

which shows faster result on total time.

505803727-a1c139af-a4aa-429a-a1c1-676b2bb32bd4

PR #1125, opened 2025-10-26
fix: Replace args with VariadicPlaceholder should remove trailing comma

samsonasik:
@TomasVotruba this PR add test and fix for rector case:

- https://github.com/rectorphp/rector-src/pull/7551

which currently we tweak via token there, that would be great if can be fixed in php-parser printer itself.

Ref https://3v4l.org/p6XC3


samsonasik:
@nikic @TomasVotruba I've applied fix 👍 Ready for review/merge 👍

samsonasik:
@TomasVotruba I've moved test under `test/code/formatPreservation` directory https://github.com/nikic/PHP-Parser/pull/1125/commits/5e60691ef671a0299e1a63e9deffe08ee7ef7403

The patch seems still needs on token level tho as far as I can understand.

PR #1126, opened 2025-10-27
Fix variadic trailing comma printer in format preserving printer - option #2

TomasVotruba:
This is alternative option to https://github.com/nikic/PHP-Parser/pull/1125
that does not modify `TokenStream`, a value object that should be read-only IMO.

To run just this test:

```bash
vendor/bin/phpunit test/PhpParser/PrettyPrinterTest.php --filter "args_to"
```

@nikic Let us know whichever you prefer or if you see a better way to fix this bug :+1:

### DIFF
--- a/lib/PhpParser/PrettyPrinterAbstract.php
+++ b/lib/PhpParser/PrettyPrinterAbstract.php
@@ -767,7 +767,7 @@ abstract class PrettyPrinterAbstract implements PrettyPrinter {
 
         $tokenCode = $this->origTokens->getTokenCode($pos, $endPos + 1, $indentAdjustment);
         if ($node instanceof Expr\CallLike && $node->isFirstClassCallable()) {
-            $tokenCode = trim($tokenCode, ',');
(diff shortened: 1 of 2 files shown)


lib/PhpParser/PrettyPrinterAbstract.php
```diff
@@ -765,7 +765,13 @@ protected function p(
             $pos = $subEndPos + 1;
         }
 
-        $result .= $this->origTokens->getTokenCode($pos, $endPos + 1, $indentAdjustment);
+        $tokenCode = $this->origTokens->getTokenCode($pos, $endPos + 1, $indentAdjustment);
+        if ($node instanceof Expr\CallLike && $node->isFirstClassCallable()) {
+            $tokenCode = str_replace(',', '', $tokenCode);
```
samsonasik: this can use `preg_replace()` with count arg = 1 to ensure only first `,` found that removed, see

https://3v4l.org/uZb6m

```suggestion
            $tokenCode = preg_replace('/,/', '', $tokenCode, 1);
```

PR #1133, opened 2025-12-01
Use canonical cast name to avoid deprecation notice on PHP >= 8.5

sebastianbergmann:


staabm:
I wonder why the 8.5 CI build did not catch this problem

sebastianbergmann:
> I wonder why the 8.5 CI build did not catch this problem

I do not see PHP 8.5 (or PHP 8.6) here: https://github.com/nikic/PHP-Parser/actions/runs/19814670264

PR #1134, opened 2025-12-06
Add arg on new without parentheses should add ) close

samsonasik:
@TomasVotruba this is detected on our rector-symfony rule:

- https://github.com/rectorphp/rector-symfony/pull/897

```diff
-new DateTime('now');
+new DateTime('now';
```

can be run with:

```diff
vendor/bin/phpunit test/PhpParser/PrettyPrinterTest.php --filter "addArgNewWithoutParentheses" 
PHPUnit 9.6.29 by Sebastian Bergmann and contributors.

F                                                                   1 / 1 (100%)

Time: 00:00.007, Memory: 12.00 MB

There was 1 failure:

1) PhpParser\PrettyPrinterTest::testFormatPreservingPrint with data set "Users/samsonasik/www/PHP-Parser/test/code/formatPreservation/addArgNewWithoutParentheses.test#0" ('Add Arg New Without Parenthes....test)', '<?php\nnew DateTime;', '$stmts[0]->expr->args = [new ...w'))];', '<?php\nnew DateTime('now');', null)
Add Arg New Without Parentheses (/Users/samsonasik/www/PHP-Parser/test/code/formatPreservation/addArgNewWithoutParentheses.test)
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
 '<?php
-new DateTime('now');'
+new DateTime'now'<?php
+new DateTime;'

/Users/samsonasik/www/PHP-Parser/test/PhpParser/PrettyPrinterTest.php:236

FAILURES!
Tests: 1, Assertions: 1, Failures: 1.
```

that only green when using `()` on the very first.

which new without parentheses got invalid diff, and we tricked to set origNode null there.

This test for it, hopefully can be fixed here instead.

### DIFF
--- a/test/code/formatPreservation/addArgNewWithoutParentheses.test
+++ b/test/code/formatPreservation/addArgNewWithoutParentheses.test
@@ -1,0 +1,9 @@
+Add Arg New Without Parentheses
+-----
+<?php
+new DateTime;
+-----
+$stmts[0]->expr->args = [new Node\Arg(new Scalar\String_('now'))];
+-----
+<?php
+new DateTime('now');


TomasVotruba:
@samsonasik Intersting. I'm almost certain this must have worked before.
Which version is the first to introduce this bug?

samsonasik:
It never working before iirc, when you have `new DateTime()` with parentheses, add `[new Arg()]` will working ok, but when we have `new DateTime` without parentheses, then add the arg, it become lost the `)` closing parentheses.

```
new DateTime() -> add Arg -> new DateTime('now') ✅ works
new DateTime -> add Arg -> new DateTime('now' ❌  not works
```

nikic:
Thanks for the report! The way insertion into an empty argument list usually works is by finding the next `(` and inserting after it. Here no `(` will be found and the code will end up doing complete nonsense.

nikic:
I believe https://github.com/nikic/PHP-Parser/commit/8c360e27327c8bd29e1c57721574709d0d706118 should fix the issue.

samsonasik:
@nikic yes, that's works, thank you. Btw, could you look on my and @TomasVotruba PRs if you have a chance, thank you.

TomasVotruba:
Thank you :+1:

PR #1136, opened 2025-12-13
Do not try to re-clone already has origNode on CloningVisitor

samsonasik:
On `CloningVisitor`, it always try to clone itself, whenever it already has `origNode` or not. 

This PR try to not re-clone when it already has an `origNode` attribute, whenever the value is set, it should already no more clone.

I reproduce the re-clone on rector usage when running unit test, it got test that shows it double cloned on rector-src /cc @TomasVotruba

### DIFF
--- a/lib/PhpParser/NodeVisitor/CloningVisitor.php
+++ b/lib/PhpParser/NodeVisitor/CloningVisitor.php
@@ -12,6 +12,10 @@ use PhpParser\NodeVisitorAbstract;
  */
 class CloningVisitor extends NodeVisitorAbstract {
     public function enterNode(Node $origNode) {
+        if ($origNode->hasAttribute('origNode')) {
+            return null;
+        }
+
         $node = clone $origNode;
         $node->setAttribute('origNode', $origNode);
         return $node;


samsonasik:
Ready for review 👍

nikic:
So rector is running CloningVisitor multiple times on the same AST? What's the reason for that?

samsonasik:
It just happen on a 1 unit test in rector-src tho (not verify other packages). I guess this may happen when multiple rules apply in a loop to revisit again until no more change.

https://github.com/rectorphp/rector-src/blob/a046e2d2f1c3f688e3ddfe214e600e73e2472344/src/Application/FileProcessor.php#L57-L79

in every iterative, the `CloningVisitor` always re-executed internally because printing is needed to verify the flag that file has been changed.

staabm:
Adding an additional functional call on every node for a rector specific problem sounds like the wrong fix. 

I think either rector should copy and adapt this visitor or fix it in a different way

samsonasik:
@staabm Okay, closing then ;)

PR #1138, opened 2026-01-28
Remove assert() on CallLike::getArgs(), returns empty array on first class callable

samsonasik:
In tools like `Rector` and `PHPStan`, we use `getArgs()` a lot, which it will trigger assert error:

```
There was 1 failure:

1) Rector\Tests\Php73\Rector\FuncCall\SetcookieRector\SetCookieRectorTest::test#1 with data ('/Users/samsonasik/www/rector-...hp.inc')
assert(!$this->isFirstClassCallable())

/Users/samsonasik/www/rector-src/vendor/nikic/php-parser/lib/PhpParser/Node/Expr/CallLike.php:32
```

when enabling:

```
ini-values: zend.assertions=1
```

I think to avoid repetitive check:

```php
        if ($node->isFirstClassCallable()) {
              return null;
        }

       $args = $node->getArgs();
```

We can carefully check that, but sometime, we forgot.

I think the returns can be early return empty array when it is first class callable.

User that want to get both `Arg` and `VariadicPlaceholder` can use `->args` public property instead.

/cc @TomasVotruba @ondrejmirtes @ruudk @staabm

### DIFF
--- a/lib/PhpParser/Node/Expr/CallLike.php
+++ b/lib/PhpParser/Node/Expr/CallLike.php
@@ -24,12 +24,14 @@ abstract class CallLike extends Expr {
     }
 
     /**
-     * Assert that this is not a first-class callable and return only ordinary Args.
+     * Return only ordinary Args.
      *
      * @return Arg[]
      */
     public function getArgs(): array {
-        assert(!$this->isFirstClassCallable());
+        if ($this->isFirstClassCallable()) {
+            return [];
+        }
         return $this->getRawArgs();
     }
 


samsonasik:
Ready for review 👍

TomasVotruba:
This would be great improvement :+1:  as the expectation would be code-dependent, not server-dependent.

ondrejmirtes:
I disagree, this would be a BC break and a source of hard-to-find bugs.

ondrejmirtes:
Also, we need to think about the design of Partial Function Application calls in the AST for PHP 8.6.

samsonasik:
@ondrejmirtes the signature doesn't change, it returns `@return Arg[]`, if assert is used, and assert php.ini is disabled, previously can returns both `Arg[]` and `VariadicPlaceholder[]`:

```php
ini_set('zend.assertions', '0');

assert(true === false);
echo "here executed";
```

which tool may can't control as can happen on `php.ini` and may be used for Arg only.

This make consistent, all Arg, or empty.

I am open for optional parameter flag, but that seems will be error incompatible param on child of ClassLike on custom Node.

nikic:
Esp. when considering PFA support (https://github.com/nikic/PHP-Parser/pull/1159), I don't think this is a good idea. For PFA, we'd have to filter out the non-variadic placeholder arguments as well, and in that case the meaning of the result would really change (as the arguments would be at different positions).

Unfortunately, PFA is something one has to explicitly deal with, it can't just be ignored :(

PR #1150, opened 2026-07-02
Add missing label characters in PrettyPrinterAbstract::$labelCharMap

TimWolla:
The underscore was missing from the list. The fix borrows the existing regular expression in `NullsafeTokenEmulator`.

As a side-effect, the only use of the ext/ctype extension has been removed, allowing the removal of the dependency on this (technically optional) extension.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,8 +15,7 @@
     "require": {
         "php": ">=7.4",
         "ext-tokenizer": "*",
-        "ext-json": "*",
-        "ext-ctype": "*"
+        "ext-json": "*"
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


TimWolla:
PHPStan failure is unrelated and already exists on master.

composer.json
```diff
@@ -15,8 +15,7 @@
     "require": {
         "php": ">=7.4",
         "ext-tokenizer": "*",
-        "ext-json": "*",
-        "ext-ctype": "*"
```
TomasVotruba, line 19: Nice work with ctype drop 👏

PR #1155, opened 2026-07-08
Fix three engine-conformance divergences in the lexer and string parser

iliaal:
Three small, independent fixes where the parser or the emulative lexer diverges from how the PHP engine tokenizes or parses the same input. In each, PHP-Parser accepts or shapes a token the engine treats differently.

## 1. `\u{...}` accepts code points above the Unicode maximum

`String_::codePointToUtf8()` accepted code points up to `0x1FFFFF` (the largest value that fits in four UTF-8 bytes), but the maximum valid Unicode code point is `0x10FFFF`. For `"\u{110000}"` through `"\u{1FFFFF}"` the parser produced a four-byte string value, while the engine rejects the whole file with `Invalid UTF-8 codepoint escape sequence: Codepoint too large`. The existing test only covered the extreme `\u{FFFFFFFFFFFFFFFF}` overflow, so this boundary slipped through. Lowered the bound to `0x10FFFF`.

## 2. Comment between `enum` and the enum name

`EnumTokenEmulator::isKeywordContext()` recognised `enum` as a keyword only when it was immediately followed by whitespace and then the `T_STRING` name. A comment in between, such as `enum/*c*/Foo` or `enum // c` followed by a newline and the name, left `enum` as a plain identifier while the engine tokenizes it as `T_ENUM`. On an older host emulating 8.1+, that turned a valid enum declaration into a parse error. Now the lookahead skips whitespace and comments before checking for the name, matching the engine (verified against `token_get_all` for the comment, line-comment, `enum = 1`, and `enum::X` cases) and the comment handling `KeywordEmulator` already applies to the preceding token.

## 3. Void cast whose whitespace spans a newline

`VoidCastEmulator::emulate()` gated the interior whitespace of a `(void)` cast with an unanchored `/[ \t]+/`, which matches any run that merely contains a space or tab even when it also contains a newline. The engine forbids newlines inside a cast, so a run like `( \nvoid)` (when a real `(void)` elsewhere in the file arms the emulator) was wrongly merged into a single `T_VOID_CAST`. Anchored the check with `\A[ \t]+\z` so only pure space/tab runs qualify.

### DIFF
--- a/lib/PhpParser/Lexer/TokenEmulator/VoidCastEmulator.php
+++ b/lib/PhpParser/Lexer/TokenEmulator/VoidCastEmulator.php
@@ -24,7 +24,7 @@ class VoidCastEmulator extends TokenEmulator {
             $numTokens = 1;
             $text = '(';
             $j = $i + 1;
-            if ($j < $c && $tokens[$j]->id === \T_WHITESPACE && preg_match('/[ \t]+/', $tokens[$j]->text)) {
+            if ($j < $c && $tokens[$j]->id === \T_WHITESPACE && preg_match('/\A[ \t]+\z/', $tokens[$j]->text)) {
                 $text .= $tokens[$j]->text;
                 $numTokens++;
                 $j++;
@@ -37,7 +37,7 @@ class VoidCastEmulator extends TokenEmulator {
             $text .= $tokens[$j]->text;
             $numTokens++;
             $k = $j + 1;
-            if ($k < $c && $tokens[$k]->id === \T_WHITESPACE && preg_match('/[ \t]+/', $tokens[$k]->text)) {
+            if ($k < $c && $tokens[$k]->id === \T_WHITESPACE && preg_match('/\A[ \t]+\z/', $tokens[$k]->text)) {
                 $text .= $tokens[$k]->text;
                 $numTokens++;
                 $k++;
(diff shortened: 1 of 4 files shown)


lib/PhpParser/Lexer/TokenEmulator/VoidCastEmulator.php
```diff
@@ -24,7 +24,7 @@ public function emulate(string $code, array $tokens): array {
             $numTokens = 1;
             $text = '(';
             $j = $i + 1;
-            if ($j < $c && $tokens[$j]->id === \T_WHITESPACE && preg_match('/[ \t]+/', $tokens[$j]->text)) {
+            if ($j < $c && $tokens[$j]->id === \T_WHITESPACE && preg_match('/\A[ \t]+\z/', $tokens[$j]->text)) {
```
nikic, line 27:
```suggestion
            if ($j < $c && $tokens[$j]->id === \T_WHITESPACE && preg_match('/^[ \t]+$/', $tokens[$j]->text)) {
```

nikic (commented):
Can you drop the enum part from this PR? It fails tests. The rest looks reasonable. (And update the commit message accordingly. I won't be able to squash merge as the PR contains multiple changes.)

iliaal:
Dropped the enum commit and updated the message; the comment-between-`enum`-and-name case only tokenizes as `T_ENUM` on 8.3+, so it isn't a clean cross-version conformance fix. The two remaining fixes are separate commits (unicode bound, void cast newline).

lib/PhpParser/Lexer/TokenEmulator/VoidCastEmulator.php
iliaal: `^[ \t]+$` matches a trailing newline (`preg_match('/^[ \t]+$/', " \n")` is 1), which is the `( \nvoid)` case this guards, so I kept `\A[ \t]+\z`. `^[ \t]+$/D` would also work.

PR #1159, opened 2026-07-18
Implement support for partial function application

### LINKED ISSUE #1158: PHP 8.6: Partial function application
PHP 8.6 implements the [partial function application RFC](https://wiki.php.net/rfc/partial_function_application_v2).

This introduces new call-site syntax: placeholder arguments such as `f(?)` and `f(1, ?)`, named placeholders such as `f(name: ?)`, and the variadic placeholder combined with bound arguments such as `f(1, ...)`, which generalizes the existing first-class callable syntax.

PHP-Parser currently rejects this syntax:

```
$ php -r 'require "vendor/autoload.php";
          (new PhpParser\ParserFactory)->createForNewestSupportedVersion()
              ->parse("<?php \$f = strlen(?);");'
PhpParser\Error: Syntax error, unexpected '?' on line 1
```

## Impact on PHPUnit

PHPUnit itself is not affected: test execution, including the test double functionality, works on PHP 8.6 without changes. The blocker is code coverage.

phpunit/php-code-coverage uses PHP-Parser to statically analyse each file that is covered: it determines executable lines, finds code units, and computes cyclomatic complexity (via sebastian/complexity) as well as size metrics (via sebastian/lines-of-code). A parse error in any one of these steps is not recoverable: as soon as a single source file that uses partial function application is included in code coverage, coverage collection aborts with an exception.

In practice this means that a project cannot adopt partial function application without giving up code coverage entirely; not just for the affected file, but for the whole test run. (At least not until https://github.com/sebastianbergmann/php-code-coverage/issues/1259 is implemented.)

## Notes

Partial function application does not introduce new lexer tokens; the call-site syntax reuses `?` and `...`. Supporting it should therefore be a matter of grammar rules and an AST node for placeholder arguments, similar in spirit to the existing `VariadicPlaceholder` node for first-class callable syntax.

PHP 8.6 is scheduled for release in November 2026. A PHP-Parser release that can parse this syntax before then would allow the PHPUnit ecosystem to raise its version constraints in time. I am happy to test development versions against PHPUnit's test suite and a PHP 8.6 development build.

sebastianbergmann:
Closes #1158.


lib/PhpParser/Node/Expr/CallLike.php
```diff
@@ -24,12 +24,30 @@ public function isFirstClassCallable(): bool {
     }
 
     /**
-     * Assert that this is not a first-class callable and return only ordinary Args.
+     * Returns whether this call expression is a partial function application, i.e. whether its
+     * argument list contains one or more "?" placeholders, or a "..." placeholder combined with
+     * other arguments.
+     */
+    public function isPartialFunctionApplication(): bool {
+        if ($this->isFirstClassCallable()) {
```
nikic: Hm, wouldn't it make more sense to treat first-class callables as a special type of partial function application? Otherwise users will always have to check both predicates.

lib/PhpParser/Node/Expr/CallLike.php
```diff
@@ -45,7 +63,7 @@ public function getArg(string $name, int $position): ?Arg {
             return null;
         }
         foreach ($this->getRawArgs() as $i => $arg) {
-            if ($arg->unpack) {
+            if (!$arg instanceof Arg || $arg->unpack) {
```
nikic: I think you'd rather want to return null after the first variadic placeholder, as you don't know the argument number anymore after that. And drop the isFirstClassCallable() check in that case.

lib/PhpParser/Node/Expr/Placeholder.php
```diff
@@ -0,0 +1,28 @@
...
+
+namespace PhpParser\Node\Expr;
+
+use PhpParser\Node\Expr;
+
+/**
+ * Represents the "?" argument placeholder of the partial function application syntax,
+ * e.g. the "?" in "foo(?)". It can only occur as the value of an Arg node.
+ */
+class Placeholder extends Expr {
```
nikic: This is not really an expression, because it can't be used in other places expressions can be used. But maybe it's best to fudge it, otherwise the types becomes annoying.

nikic:
cc @ondrejmirtes on representation.

lib/PhpParser/Node/Expr/CallLike.php
sebastianbergmann: Implemented in cb86a570129468c6fc6c5af3cbe3184911953340.

lib/PhpParser/Node/Expr/CallLike.php
sebastianbergmann: Implemented in 9fb848c3bb1a1513c89851ab5341c9ac08758ec1.

ondrejmirtes:
Yeah, Placeholder being Expr is annoying.

Why the Placeholder can't be the same thing as VariadicPlaceholder?

This code:

```php
if ($arg instanceof VariadicPlaceholder || $arg->value instanceof Placeholder) {
```

could become:

```php
if ($arg instanceof VariadicPlaceholder || $arg instanceof Placeholder) {
```

Is it just because of named arguments or maybe other properties `Arg` has? Placeholder could have the same property (properties).

sebastianbergmann:
@ondrejmirtes I hope I understood correctly what you were asking for; I implemented it in 31754522a5919d2c10422907ba609d49eb3e7999.

ondrejmirtes:
Yeah, thank you!

nikic:
Would it be valuable for the two placeholders to have a common base class?

sebastianbergmann:
> Would it be valuable for the two placeholders to have a common base class?

I do not think so: the two placeholders share no sub-nodes (`Placeholder` has `name`, `VariadicPlaceholder` has none) and no behavior. Its only users inside the library would be the `instanceof` check in `isPartialFunctionApplication()` and a slightly shorter type union in a few docblocks.

Should a real use case show up (in PHPStan, for example), an empty abstract base or marker interface can be added later without breaking anything, since `VariadicPlaceholder` is not final.

ondrejmirtes:
No need for PHPStan either. I use `getArgs()` everywhere. If a FuncCall is first-class callable, I don't run FuncCall rules for it (because they were written before first-class callables were a thing and most people do not think about first-class callables when they write a rule for FuncCall). Instead, I have virtual nodes such as [FunctionCallableNode](https://apiref.phpstan.org/2.2.x/PHPStan.Node.FunctionCallableNode.html) for first-class callables.

So I plan to do the same thing for partial applications. I don't know yet whether I will be able to make first-class callable a specialization of partial application call, I will have to check the RFC in depth.

nikic:
This looks good to me. My only remaining question is whether Placeholder should be ArgPlaceholder to be more specific? If PHP introduces placeholders in other places, they probably couldn't use this class (because the `name` concept is argument-specific).

sebastianbergmann:
> My only remaining question is whether Placeholder should be ArgPlaceholder to be more specific?

I implemented this in f69d2a32134532c3a43c5904b840d3d3c26a940e last week.

nikic (approved):

PR #1160, opened 2026-07-23
Cheap check first in `dereferenceLhsRequiresParens()`

staabm:
do simple bool comparison before going the more expensive route via another method call

born from a PHPStan case in which pretty printing of a files expressions turn into the main bottleneck
grafik

repro:
run phpstan on https://github.com/phpstan/phpstan-src/blob/7e97b9e67a506cd056ed3f34ad739a388c6d7942/tests/bench/data/nullsafe-chain-walk.php like 
`phpstan analyze nullsafe-chain-walk.php --debug`

----

this PR does not resolve this bottleneck though


nikic:
This is only going to make a positive difference if the value is actually a constant (i.e. you are printing `FOO[0]`), which is a really rare case. So I would not expect this to be beneficial in practice.

staabm:
while I agree its a bit artificial, the fix itself is straitforward (and maybe not worth discussing?)

PR #1162, opened 2026-08-09
Fix inclusive endFilePos in unterminated comment error range

lenamonj:
`doc/component/Lexer.markdown` defines `endFilePos` as "Offset into the code string of the last character that is part of the node". The unterminated-comment error passes `Token::getEndPos()`, which is `$this->pos + \strlen($this->text)` - one past the last byte.

The reported end column is therefore outside the input:

```
'<?php /*'              8 bytes -> Unterminated comment from 1:7 to 1:9
'<?php /*\n'            9 bytes -> Unterminated comment from 1:7 to 2:1
'<?php $x = 1; /* abc' 20 bytes -> Unterminated comment from 1:15 to 1:21
```

An 8-byte input has no column 9, and the second case reports a position on line 2 of a file whose line 2 is empty.

Every other error site writes this attribute inclusively - `Lexer.php:56` passes `$token->pos`, `ParserAbstract.php:497` uses `$afterEndToken->pos - 1`, `ParserAbstract.php:519` passes `$token->pos`. This is the one site that does not.

`getEndPos() - 1` alone is not enough, because the comment token includes its trailing newline when the input ends in one:

```
bare getEndPos()-1 -> byte 8 = '\n', column 9 on a line of 8 characters
with rtrim         -> byte 7 = '*',  column 8
```

A position on a line terminator has no meaningful column, so the range stops at the last content character and `endLine` is derived from the same text.

## Verification

`LexerTest::testError` already covers these inputs, so no new test is added. Three of its expectations recorded the old values, which is why the suite was green over this. Both halves are load-bearing, proven on `master`:

| tree | result |
|---|---|
| `master`, untouched | OK (1895 tests, 2766 assertions) |
| fix only, expectations left alone | 8 failures |
| expectations only, fix reverted | 8 failures |
| both | OK (1895 tests, 2766 assertions) |

```
1) PhpParser\LexerTest::testError with data set #0
   ('<?php /*', array('Unterminated comment from 1:7 to 1:9'))
-'Unterminated comment from 1:7 to 1:9'
+'Unterminated comment from 1:7 to 1:8'
```

`test_old/run-php-src.sh 7.4.33` exits 0 over 14,695 files with no format-preserving mismatches, on PHP 8.5.4.

### DIFF
--- a/lib/PhpParser/Lexer.php
+++ b/lib/PhpParser/Lexer.php
@@ -102,11 +102,16 @@ class Lexer {
         // Check for unterminated comment
         $lastToken = $tokens[$numTokens - 1];
         if ($this->isUnterminatedComment($lastToken)) {
+            // endFilePos is inclusive, unlike the exclusive getEndPos(). An unterminated comment
+            // is the one comment token that can end on a newline, and a range ending on a line
+            // terminator has no column, so the reported range stops at the last content
+            // character instead.
+            $text = \rtrim($lastToken->text, "\r\n");
             $errorHandler->handleError(new Error('Unterminated comment', [
                 'startLine' => $lastToken->line,
-                'endLine' => $lastToken->getEndLine(),
+                'endLine' => $lastToken->line + \substr_count($text, "\n"),
                 'startFilePos' => $lastToken->pos,
-                'endFilePos' => $lastToken->getEndPos(),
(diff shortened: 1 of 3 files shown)


lenamonj:
Friendly ping. Still clean against master; happy to adjust anything you'd like.

PR #1163, opened 2026-08-22
Render prefix operator chains in linear time

iliaal:
Pretty-printing a chain of prefix operations (`"!!!!...$a"`, nested casts, chained assignments, ...) recurses once per nesting level, and each level copies the full rendering of its operand into a new string. Print cost is therefore quadratic in the nesting depth with a large memory constant, while parsing the same input is cheap because the parser is iterative:

```text
input "!" x N        parse     print (before)   peak mem (before)
N=25_000             0.12s     0.15s            122 MB
N=50_000             0.21s     0.28s            241 MB
N=100_000            0.29s     1.09s            479 MB
N=200_000 (~200KB)   ok        fatal at 1G      --
```

This matters for anything that parses and prints untrusted or machine-generated code (formatters, codemods, refactoring tools), where nesting depth is attacker- or generator-controlled and input-length caps do not bound it well: 25KB of input already produced a 122MB print peak.

This change walks the whole chain instead, renders the innermost operand once and composes the operators around it from precomputed segments. The output is byte-identical to the recursive rendering, including precedence parentheses and the special case that avoids printing `+(+$a)` as `++$a`; I verified this with a differential harness over 4828 chain/context combinations against the previous implementation, in addition to the existing test suite.

Operations whose operator string is dynamically computed or version-dependent (arrow functions, yield) are excluded from the fast path; they terminate the chain and are rendered recursively as before.

After:

```text
input "!" x N        parse     print (after)    peak mem (after)
N=100_000            0.14s     0.05s            103 MB
N=200_000            0.30s     0.08s            203 MB
```

Scaling is now linear; mixed chains (`!(int)-` x 30000) print in 0.06s.

### DIFF
--- a/lib/PhpParser/PrettyPrinterAbstract.php
+++ b/lib/PhpParser/PrettyPrinterAbstract.php
@@ -495,17 +495,37 @@ abstract class PrettyPrinterAbstract implements PrettyPrinter {
      * This mirrors the pPrefixOp() call sites in Standard. Operations with dynamically
      * computed or version-dependent operator strings (arrow functions, yield) are excluded,
      * they simply terminate the chain and are rendered recursively.
+     *
+     * @return array{class-string, string, Node}|null
      */
     private function getChainablePrefixOp(Node $node): ?array {
-        if ($node instanceof Expr\BooleanNot) return [Expr\BooleanNot::class, '!', $node->expr];
-        if ($node instanceof Expr\BitwiseNot) return [Expr\BitwiseNot::class, '~', $node->expr];
-        if ($node instanceof Expr\UnaryMinus) return [Expr\UnaryMinus::class, '-', $node->expr];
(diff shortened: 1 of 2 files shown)


iliaal:
The 6 errors in the PHP 8.6 job are unrelated to this change: they all fail during parsing (float.test, explicitOctal.test, unicodeEscape.test hitting 'Input number is larger than PHP_INT_MAX' in Node/Scalar/Float_.php), which this diff does not touch, and CodeParsingTest never invokes the printer. That looks like a PHP 8.6 nightly behavior change since master's last green run on this workflow; happy to see it confirmed by re-running the 8.6 job on plain master.

nikic:
If you want to protect against quadratic printing on adversarial input, just addressing one instance of this problem is not useful -- you should be able to easily come up with some minor variation that happens to not involve only prefix operators and still exhibits the same problem. If we wanted to address this, we'd have to change the pretty printer architecture entirely, to not do recursive printing in the way it currently does.

PR #1164, opened 2026-08-24
perf: Skip escape sequence processing for strings without backslashes

samsonasik:
`String_::parse()` always ran `str_replace()` for single-quoted strings and `str_replace()` + `preg_replace_callback()` for double-quoted strings, even when the literal contained no backslash and there was nothing to unescape. 

This PR skips escape sequence processing for strings without backslashes.

Tested on php-parser vendor, laravel framework, and coolify, it shows faster:

Corpus | Double-quoted tokens | parse() gain (function-level, double-quoted only)
-- | -- | --
PHP-Parser vendor | 2,105 | 1.70 → 1.27 ms (~25%)
laravel-framework | 9,157 | 10.66 → 10.07 ms (~5.5%)
coolify (Laravel app + vendor) | 148,958 | 166.6 → 164.5 ms (~1.3%)

### DIFF
--- a/lib/PhpParser/Node/Scalar/String_.php
+++ b/lib/PhpParser/Node/Scalar/String_.php
@@ -77,12 +77,12 @@ class String_ extends Scalar {
         $quote = $str[$bLength];
         $str = substr($str, $bLength + 1, -1);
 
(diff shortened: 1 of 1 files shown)


samsonasik:
Ready to merge 👍

PR #1166, opened 2026-09-06
Handle `resource` types in builder APIs

staabm:


staabm:
Doesn't make sense, resource can't be a native typehint:

https://3v4l.org/h3QkW

